Issue #1, opened 2018-05-30
Rethink drop-down menu, particularly chevrons


mor10:
## Issue Overview
Drop-down menu doesn't work quire right. Chevron shows up where it shouldn't, has minimal functionality. Could use better CSS and possibly retire the whole JS solution.

mor10:
My idea:

Use `focus-in` for mouse and keyboard nav (as is the case now) and use the chevrons _only_ for touch-interactions. Possibly apply the `screen-reader-text` class to the chevron buttons so they are skipped over by keyboard users since they don't do anything. 

Doing this would eliminate the need for most of the JS and leave us with a very simple function. No need for the string passing and other complex behavior.

mor10:
closed via pull request above. good work @bamadesigner

PR #4, opened 2018-06-11
Optimizing main menu to be accessible

bamadesigner:
## Description
A first take at issue #1. This should address keyboard, touch, and mouse. Allows keyboard users to skip over large submenus, clicking to access what they want.

It's written in a way that still works gracefully if no JS. And I cleaned up the chevrons so they're icons now, instead of HTML entities.

Also done:

- Optimized the way we were adding the dropdown markup. We can use the "nav_menu_link_attributes" filter, which removes the need for regex.
- Added an aria-label for the <nav> and for the nav toggle.
- Added "aria-current" to the current menu item.

Addresses issue #
#1 

## List of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [X] This pull request relates to a ticket.
- [X] My code is tested.

### DIFF
--- a/dev/config/cssVariables.json
+++ b/dev/config/cssVariables.json
@@ -1,16 +1,17 @@
 {
 	"variables": {
 		"global-font-color": "#333",
-    "global-font-family": "'Crimson Text', serif",
-    "global-font-size": "20",
-    "global-font-line-height": "1.4",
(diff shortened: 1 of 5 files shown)


mor10:
excellent solution. 10 points.

Issue #6, opened 2018-06-14
Enhancement: Support HTTPS


miklb:
## Issue Overview

Currently Browser sync only supports http
## Describe your environment

local dev environment running nginx via hombrew in macOS. Using a local SSL cert (not self-signed)
## Possible Solution

```
gulp.task('browser-sync', function() {
    browserSync.init({
        proxy: projectURL,
        https: {
            key: keyPath,
            cert: certPath
        }
    })
})
```
where `keyPath` and `certPath` are user variables for the respective paths. `projectURL` I believe is same as current `proxyURL` in the config. 

I'd guess a simple check for those variables from `themeConfig.js` and then passing them into the `https` key for `browserSync` should work.

Does that sound logical and if so, I can take a stab at a PR.

miklb:
Not quite that simple. Something in the watch function is failing, but the error message is unhelpful at this point. Will continue digging.

miklb:
As long as I don't pass the port with `https` I have it working.

bamadesigner:
This issue has been resolved and merged in v1.1 (which is under development and will be the next major release). To test this feature, you can check out the v1.1. branch. Thanks!

hillal20:
![Image](https://github.com/user-attachments/assets/6926abbf-6cbf-400e-97ca-a12fc782c9c2)

hillal20:
The problem is still existing until 2025

PR #7, opened 2018-06-15
add: support for https in Browser Sync

miklb:
address #6

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -82,13 +82,25 @@ const server = browserSync.create();
 // Initialize the BrowserSync server conditionally:
 function serve(done) {
 	if (config.dev.browserSync.live) {
-		server.init({
+		var serverConfig = {
 			proxy: config.dev.browserSync.proxyURL,
-			port: config.dev.browserSync.bypassPort,
 			liveReload: true
-		});
+		};
+		if (config.dev.browserSync.https) {
+			var httpConfig = {
+				https: {
+					key: config.dev.browserSync.keyPath,
+					cert: config.dev.browserSync.certPath
+				}
+			}
+		} else {
+			var httpConfig = {
+				proxy: config.dev.browserSync.proxyURL
+			}
+		}
 	}
-	done();
+	Object.assign(serverConfig, httpConfig);
+	server.init(serverConfig);
 }
 
 // Reload the live site:
--- a/dev/config/themeConfig.js
+++ b/dev/config/themeConfig.js
@@ -9,7 +9,12 @@ module.exports = {
 		browserSync: {
 			live: true,
 			proxyURL: 'wprig.test:8888',
-			bypassPort: '8181'
+			bypassPort: '8181',
+			https: false,
+
+			// If true, path to local cert & key.
+			certPath: '',
+			keyPath: ''
 		},
 		browserslist: [ // See https://github.com/browserslist/browserslist
 			'> 1%',


bamadesigner:
Hi @miklb! Thanks for the PR. Do you mind removing all of the indent changes so we can keep the PR limited to browserSync changes?

miklb:
Sure. Didn't realize I messed up indents. I'll take a look at my linter settings and clean it up.

gulpfile.babel.js
```diff
@@ -82,15 +82,26 @@ const server = browserSync.create();
 // Initialize the BrowserSync server conditionally:
 function serve(done) {
 	if (config.dev.browserSync.live) {
-		server.init({
-			proxy: config.dev.browserSync.proxyURL,
-			port: config.dev.browserSync.bypassPort,
-			liveReload: true
-		});
+		if (config.dev.browserSync.https) {
```
bamadesigner: If we need to setup different arguments, let's start with the commonalities, e.g.:
`var serverConfig = {
    proxy: config.dev.browserSync.proxyURL,
    liveReload: true
}
if (config.dev.browserSync.https) {
    // Add to serverConfig
} else {
    // Add to serverConfig
}
server.init(serverConfig);
`

miklb: OK, I'll tackle this next. Figured out it was Prettier extension in VS Code that did all of that formatting change.

miklb:
ugh, I did a poor job of backing out my changes and cleaning up the linting. Let me know if you would prefer a fresh PR.

Thanks for the feedback.

bamadesigner:
Hey @miklb! Thanks for the cleanup!

I don't think we need a new PR but would you mind squashing your commits? If you've never squashed before, just let me now and I can walk you through it. It can be confusing.

miklb:
It’s been a while but I’m confident I can figure it out. Thanks again for walking me through contributing

miklb:
did that work?

bamadesigner:
@miklb No. If it was squashed, it would be only 1 commit.

If you're familiar with Terminal, you can do the following:

1. Make sure your branch is up to date with master.
2. Run `git rebase -i master`.
3. You should see a list of commits, starting with `pick`.
4. Make sure the first one says `pick` and change the rest from `pick` to `squash`. Basically, this means you want to squash your commits into the one before it.
5. Save and close the editor.
6. Then it will give you the opportunity to change the commit message. Save and close again.
7. Then you have to force push the new squashed commit: `git push -f`

I hope this was helpful. It's tricky but once you figure it out, it's really helpful.

miklb:
ah, forced push is what I was missing. Thanks! I'll fix.

mor10:
@bamadesigner That rundown of how to squash commits should be added to the Wiki for future reference!

mor10:
@miklb Something about this PR makes the watch tasks in my test environment not work. I am too jetlagged to see what's going on here, but there must be something happening in the gulpfile. To test, simply use this PR, `npm run build`, and make a change to a CSS or PHP file. In my setup, nothing happens, but if I revert to a different commit, it works fine.

@bamadesigner can you confirm?

miklb:
I'll look closer, I was having a difficult time testing locally with the defaults, thus the PR.

bamadesigner:
Thanks for the PR and the hard work @miklb! You were the foundation/inspiration for us to merge #57 and add https support for BrowserSync!

PR #9, opened 2018-06-15
Ensuring main menu works for page menus

bamadesigner:
## Description
After testing the main menu some more, I realized the accessibility additions for the main menu didn't work if a menu was not assigned to the "primary" menu location. They didn't work because WordPress then prints a page menu instead of a nav menu and the container markup is different.

These changes ensure the markup is the same and all of the functionality works.

The only front-end markup that changes is the outer container went from being the default `<div class="menu-main-menu-container">` to a custom class I created: `<div class="primary-menu-container">`.

I left a @TODO note to remark that the dropdown symbols do not get added for page menus if JS is disabled because the equivalent filter that we use for nav menus (walker_nav_menu_start_el) does not exist for page menus.

I also had to tweak how `aria-current` is added to make sure it covers both nav menu post objects and and other post objects.

## List of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] My code is tested.

### DIFF
--- a/dev/header.php
+++ b/dev/header.php
@@ -71,13 +71,18 @@
 					<?php esc_html_e( 'Menu', 'wprig' ); ?>
 				</button>
 
-				<?php
-				wp_nav_menu(
-					array(
-						'theme_location' => 'primary',
-						'menu_id'        => 'primary-menu',
-					)
-				);
-				?>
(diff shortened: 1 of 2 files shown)


mor10:
Merging this to development branch.

PR #21, opened 2018-06-17
Ensure theme slug and theme name are replaced independently

### LINKED ISSUE #18: Handle replacement of theme name separately from the slug
## Issue Overview
By default, the current codebase uses `wprig` as both the theme slug and the theme name. This is however an unlikely scenario for most themes (consider `twentyseventeen` vs `Twenty Seventeen` for example), so these two should be handled separately when making replacements.

## Steps to Reproduce
1. Set the theme name to a proper theme name, such as a string containing spaces and uppercase/lowercase.
2. Run `gulp build`.
3. The code will not work since function names suddenly look like `My Theme_setup()` for example.

## Expected Behavior
Theme name and slug should be two separate things. By default `wprig` should be the slug and `WP Rig` the name.

## Possible Solution
A new key `theme.slug` should be added to the config in `dev/config/themeConfig.js`, with the value `wprig`. The `theme.name` value should be changed to `WP Rig`. All occurrences where `wprig` is actually used as theme name should be manually replaced with `WP Rig` too. Then, in the gulp scripts, always replace both values appropriately.

felixarntz:
## Description
Addresses issue #18 

This PR separates the theme name from the theme slug and thus allows for those to work correctly for themes that do not use the same value for both (which is by far the majority).

By having a `theme.slug` with default value "wprig" and a `theme.name` with default value "WP Rig", one can easily change the two values individually. While this requires a developer to change two values instead of just one, I think it makes sense here. The replacements are then made with both values individually, per the tweaks to the gulpfile.

Closes #18.

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/dev/config/themeConfig.js
+++ b/dev/config/themeConfig.js
@@ -2,7 +2,8 @@
 
 module.exports = {
 	theme: {
-		name: 'wprig',
+		slug: 'wprig',
+		name: 'WP Rig',
 		author: 'Morten Rand-Hendriksen'
 	},
 	dev: {
--- a/dev/footer.php
+++ b/dev/footer.php
@@ -22,7 +22,7 @@
 		<span class="sep"> | </span>
 		<?php
 			/* translators: 1: Theme name, 2: Theme author. */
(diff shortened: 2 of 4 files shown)


felixarntz:
Not sure what the test failure is about, it should work after restarting the respective job.

bamadesigner:
@felixarntz I restarted it and it passed 👍

mor10:
Yes to this. Will merge tomorrow.

Issue #30, opened 2018-06-25
Re-running npm run build does not rebuild the PHP files


hellofromtonya:
## Issue Overview

After you've run `npm run build`, running it again after changing the [slug in `themeConfig.js`](https://github.com/wprig/wprig/blob/master/dev/config/themeConfig.js#L5) does not rebuild the PHP files.

Why is this a problem?

Imagine that a developer makes a typo in the slug or wants to change its name.  The _assumption_ is the theme is regenerated.  But that's not completely true.  Instead, the developer is left to either search/replace to rename or delete all of the theme files and start again.  Bummer.

@ygagnon notes this problem in issue #29 where he named the theme's slug `wp` and then needed to change it.  It will happen for other developers too.  Shoot, it'll likely happen for all of us where at some point we want to rename.

## Steps to Reproduce

1. Clone wprig and then navigate into the folder.
2. Open `wprig/dev/config/themeConfig.js`.
3. Change the theme's slug.
4. Save the file.
5. In terminal, do: `npm run build`.  Everything is cool and happy.
6. Repeat steps 2-5.

Whoops, the PHP prefixes and DocBlocks are not updated.

## Expected Behavior

Regenerate the theme include the PHP files.

## Current Behavior

Re-running `npm run build` after changing the theme's slug do the following:

```
T-iMac:hello hellofromtonya$ npm run build
> wprig@1.0.0 build /Users/hellofromtonya/Sites/sandbox/app/public/wp-content/themes/hello
> gulp

[15:54:21] Failed to load external module @babel/register
[15:54:21] Requiring external module babel-register
[15:54:23] Using gulpfile ~/Sites/sandbox/app/public/wp-content/themes/hello/gulpfile.babel.js
[15:54:23] Starting 'default'...
[15:54:23] Starting 'php'...
[15:54:23] Finished 'php' after 119 ms
[15:54:23] Starting 'scripts'...
[15:54:23] Starting 'jsMin'...
[15:54:23] Starting 'jsLibs'...
[15:54:23] Finished 'jsLibs' after 11 ms
[15:54:23] Finished 'jsMin' after 18 ms
[15:54:23] Finished 'scripts' after 20 ms
[15:54:23] Starting 'sassStyles'...
[15:54:23] Finished 'sassStyles' after 5.54 ms
[15:54:23] Starting 'styles'...
[15:54:23] dev/style.css
[15:54:23] dev/css/comments.css
[15:54:23] dev/css/content.css
[...]

PR #31, opened 2018-06-25
Run PHP task on all files when theme slug is changed in themeConfig.js.

mor10:
## Description

Addresses issue #30. 

Adds new `watch` task monitoring `themeConfig.js` for changes and runs PHP task any time the file is updated. Compares stored theme slug with theme slug in current `themeConfig.js`, and if no match is found, skips over the `newer` filter forcing all PHP files to be reprocessed so function names are rewritten with the new slug.

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -33,11 +33,13 @@ import zip from 'gulp-zip';
 
 // Import theme-specific configurations.
 var config = require('./dev/config/themeConfig.js');
+var themeSlug = config.theme.slug;
 
 // Project paths
 const paths = {
 	config: {
-		cssVars: './dev/config/cssVariables.json'
+		cssVars: './dev/config/cssVariables.json',
+		themeConfig: './dev/config/themeConfig.js'
 	},
 	php: {
 		src: ['dev/**/*.php', '!dev/optional/**/*.*'],
@@ -111,8 +113,15 @@ function reload(done) {
  */
 export function php() {
 	config = requireUncached('./dev/config/themeConfig.js');
+	// Check if theme slug has been updated.
+	let newThemeSlug = false;
+	if ( config.theme.slug !== themeSlug ) {
+		newThemeSlug = true;
+		themeSlug = config.theme.slug;
+	}
 	return gulp.src(paths.php.src)
-	.pipe(newer(paths.php.dest))
+	// If theme slug has not been updated, run task on changed files only.
+	.pipe(gulpif(!newThemeSlug, newer(paths.php.dest)))
 	.pipe(phpcs({
 		bin: 'vendor/bin/phpcs',
 		standard: 'WordPress',
@@ -124,6 +133,7 @@ export function php() {
 	.pipe(replace('WP Rig', config.theme.name))
 	.pipe(gulp.dest(paths.verbose))
 	.pipe(gulp.dest(paths.php.dest));
+
 }
 
 /**
@@ -235,6 +245,7 @@ export function images() {
  */
 export function watch() {
 	gulp.watch(paths.php.src, gulp.series(php, reload));
+	gulp.watch(paths.config.themeConfig, gulp.series(php, reload));
 	gulp.watch(paths.config.cssVars, gulp.series(styles, reload));
 	gulp.watch(paths.styles.sass, sassStyles);
 	gulp.watch(paths.styles.src, gulp.series(styles, reload));
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -33,7 +33,8 @@ import zip from 'gulp-zip';
 
 // Import theme-specific configurations.
 var config = require('./dev/config/themeConfig.js');
-var themeSlug = config.theme.slug;
+var themeConfig = config.theme;
+themeConfig.isFirstRun = true;
 
 // Project paths
 const paths = {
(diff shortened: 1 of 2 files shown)


hellofromtonya:
Using the code as is in this branch, we do not rebuild the PHP files on the first pass when the files already exist and there is one or more changes in the `themeConfig.js` file.  It works in watch mode though.

How to reproduce?

1. Do `npm run build`.
2. Then stop the watcher.
3. Change the theme's slug in `themeConfig.js`.
4. Do `npm run build`.  **Result:  PHP Files are not rebuilt.**

Whoopsie.  That doesn't fit all scenarios.

## Altering the Code

When removing the new code in the `php()` function as well as the original `.pipe(newer(paths.php.dest))`, bam, it works in all scenarios.  WooHoo!

### Test Results

Here are the tests I ran:

1. Changed the theme's slug in the `themeConfig.js`.  Then ran `npm run build`.  **Result: Works.**
2. Changed the theme's slug while in watch. **Result: Works.**
3. Changed the theme's name while in watch mode. **Result: Works.**
4. Stopped the watcher.  Changed both the theme's slug and name. Restarted. **Result: Works.**
5. Repeated 2 and 3. **Result: Works.**
6. Stopped. Changed the name only. Restarted. **Result: Works.**
7. Repeated 2 and 3. **Result: Works.**
8. Repeated the process of changing slug, name, or both.  **Result: Works.**

### New Code

@mor10 Here's the new code for the `php` function.

```
/**
 * Build the PHP files and run PHP Code Sniffer.
 */
export function php() {
	config = requireUncached('./dev/config/themeConfig.js');

	return gulp.src(paths.php.src)
	// Run PHPCS.
	.pipe(phpcs({
		bin: 'vendor/bin/phpcs',
		standard: 'WordPress',
		warningSeverity: 0
	}))
	// Log all problems that was found
	.pipe(phpcs.reporter('log'))

	// Replace the theme's slug and name in the build files.
	.pipe(replace('wprig', config.theme.slug))
	.pipe(replace('WP Rig', config.theme.name))

	// Set the build destinations.
	.pipe(gulp.dest(paths.verbose))
	.pipe(gulp.dest(paths.php.dest));
}
```

hellofromtonya:
@mor10 Here's the revised [`gulpfile.babel.js` in my copy of the branch](https://github.com/hellofromtonya/wprig/blob/try/watchthemename/gulpfile.babel.js) to help you out.

gulpfile.babel.js
```diff
@@ -111,8 +113,15 @@ function reload(done) {
...
+	// Check if theme slug has been updated.
+	let newThemeSlug = false;
+	if ( config.theme.slug !== themeSlug ) {
+		newThemeSlug = true;
+		themeSlug = config.theme.slug;
+	}
 	return gulp.src(paths.php.src)
-	.pipe(newer(paths.php.dest))
+	// If theme slug has not been updated, run task on changed files only.
+	.pipe(gulpif(!newThemeSlug, newer(paths.php.dest)))
```
hellofromtonya: We don't need this line of code as it is preventing a rebuild in the following scenario:

1. Files already exist, i.e. where previously built.
2. AND theme's slug and/or name changed in the themeConfig.js file.
3. AND when starting gulp again via `npm run build`.


Removing it fixes the above scenario.  All scenarios do cause the PHP files to rebuild. WooHoo!

gulpfile.babel.js
```diff
@@ -111,8 +113,15 @@ function reload(done) {
  */
 export function php() {
 	config = requireUncached('./dev/config/themeConfig.js');
+	// Check if theme slug has been updated.
```
hellofromtonya: Lines 116 to 121 can be removed.

gulpfile.babel.js
```diff
@@ -33,11 +33,13 @@ import zip from 'gulp-zip';
 
 // Import theme-specific configurations.
 var config = require('./dev/config/themeConfig.js');
+var themeSlug = config.theme.slug;
```
hellofromtonya: This line can be removed.

gulpfile.babel.js
mor10: The idea here is for the PHP task to only run on files it needs to run on. The newer method ensures no unnecessary builds. Removing it results in all php files being rebuilt every time a single php file is altered which is resource intensive and unnecessary.

mor10:
Off the top of my head a solution might be to not set `themeSlug` until after the PHP process has run for the first time. That would result in all PHP files being rebuilt every time the build process is started.

gulpfile.babel.js
hellofromtonya: Modify L116-121 to the following code:


```
	let isRebuild = themeConfig.isFirstRun ||
		( themeConfig.slug !== config.theme.slug ) ||
		( themeConfig.name !== config.theme.name );
	if ( isRebuild ) {
		themeConfig.slug = config.theme.slug;
        themeConfig.name = config.theme.name;
	}

	// Reset first run.
	if ( themeConfig.isFirstRun ) {
        themeConfig.isFirstRun = false;
    }
```

gulpfile.babel.js
hellofromtonya: To take into account:

1. On script load, i.e. first run.
2. On name or slug change

we can modify this line of code to be:

```
	// If not a rebuild, then run tasks on changed files only.
	.pipe(gulpif(!isRebuild, newer(paths.php.dest)))
```

gulpfile.babel.js
hellofromtonya: We want to take into account slug and name, as either or both can change.  If either changes, then we want to trigger a rebuild process.  Therefore, I'd change this line of code to:

```
var themeConfig = config.theme;
themeConfig.isFirstRun = true;
```

The advantage here is that we capture the initial theme's configuration.  We'll use this to modify it as we run the php task..

The `themeConfig.isFirstRun` property is capturing the initial "on script load" occurrence.  That way we can run the php function when the first first loads.  It's a cheap way to note that the script is first running.

hellofromtonya:
@mor10 Take a look at this possible solution. Here it's doing the following:

1. Runs the php() task on script load, i.e. 1st run.
2. Runs the rebuild tasks when the theme's slug or name (or both) change.

felixarntz:
General remark: We need to consider whether `slug` will always be the only parameter that needs to be covered here. The slug at the moment is only a single word, however what happens if the the slug would be something consisting of two words? If I want to have `my-theme`, this would only work as a textdomain replacement, but no longer as function name prefix replacement (we'd probably need to sanitize it to `my_theme` there. If the code at some point includes classes or namespaces, we would also need to account for `My_Theme`.

I'm not even sure this issue here is the right one to make that comment, but I generally wanted to highlight this. I wonder whether the starter theme would be better off using distinct `wp_rig` and `wp-rig` strings. That would make replacements in such cases easily possible while at the moment it uses the simplest use-case.

mor10:
@felixarntz can you write that last comment up in a separate issue. I will add it to the 1.1 milestone.

Issue #33, opened 2018-06-26
Improve logic in wprig_scripts()


Tabrisrp:
## Issue Overview
In `wprig_scripts()`, a conditional check for the return of `wprig_is_amp()` is done, and loads the scripts only if AMP is not active.

## Possible Solution
The logic would be better by inverting the conditional check, returning early if AMP is active, and then doing all the enqueuing.

mor10:
Merged. Closing.

PR #34, opened 2018-06-26
Improve conditional logic in wprig_scripts()

Tabrisrp:
## Description
Addresses issue #33

Inverts the conditional check in `wprig_scripts()`, returning early if AMP is active, and then doing all the enqueuing.

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/dev/functions.php
+++ b/dev/functions.php
@@ -288,25 +288,26 @@ add_action( 'wp_enqueue_scripts', 'wprig_styles' );
  */
 function wprig_scripts() {
 
-	// If the AMP plugin is not active, enqueue custom scripts.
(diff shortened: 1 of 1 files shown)


hellofromtonya:
Picture me giving you a high five, @Tabrisrp!  Way to improve the quality of this function!  

+1 for this PR. 

Just my 2 cents...using a guard clause, as you did, does a couple of things for us:

1. Clearly communicates that if the condition is not right, we're bailing out.  No misunderstandings.
2. Shifts the business logic to the left.  That makes it easier to read and understand.

felixarntz:
+1 from me as well! Should be able to be merged as is.

PR #38, opened 2018-06-28
Remove unused $post_count from index.php

Soean:
In the index.php file is a `$post_count` variable to count the posts, which is not used in the template. I don't think we need this one in the boilerplate when we don't use it.

This PR removes it.


mor10:
Oops. That is indeed junk. Thanks.

PR #39, opened 2018-06-29
add: version number

davidsword:
Add `WPRIG_VERSION` version number as a constant for theme. Makes it so while `wp_debug` is `true`, force refreshes occur with a `time()` value. Can also force refresh all resources to visitors by manually changing YYYYMMDD timestamp, so dev can do a cache bust for visitors. Reduces repetition.

---

---

- [x] My code is tested.

### DIFF
--- a/dev/functions.php
+++ b/dev/functions.php
@@ -8,6 +8,14 @@
  */
 
 /**
+ * Default version number for wp_{enqueue|register}_{script|style}.
+ *
+ * WP_DEBUG will force refresh while developing.
+ * Manualy changing YYYYMMDD value will force refresh all resources for visitors.
+ */
+define( 'WPRIG_VERSION', ( WP_DEBUG ? time() : '20180514' ) );
+
+/**
  * Sets up theme defaults and registers support for various WordPress features.
  *
  * Note that this function is hooked into the after_setup_theme hook, which
@@ -242,7 +250,7 @@ function wprig_gutenberg_styles() {
 	wp_enqueue_style( 'wprig-fonts', wprig_fonts_url(), array(), null );
 
 	// Enqueue main stylesheet.
-	wp_enqueue_style( 'wprig-base-style', get_theme_file_uri( '/css/editor-styles.css' ), array(), '20180514' );
+	wp_enqueue_style( 'wprig-base-style', get_theme_file_uri( '/css/editor-styles.css' ), array(), WPRIG_VERSION );
 }
 add_action( 'enqueue_block_editor_assets', 'wprig_gutenberg_styles' );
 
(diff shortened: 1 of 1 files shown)


mor10:
This is a great idea that will cause breaking changes. I'm putting it under the v1.1 milestone to give others time to pipe in. A 1.1 branch will be created shortly.

bamadesigner:
I like the idea of implementing a way to force refresh assets, but I'd like to consider a different approach. I'm not a fan of piling on PHP constants.

Also, what's being defined for `WPRIG_VERSION` is not actually the version number so that could be confusing.

Instead of defining the constant, what if we created a function to get the assets version. It could have a filter too, e.g.:

```
function wprig_get_asset_version() {
    return apply_filters( 'wprig_asset_version', WP_DEBUG ? time() : wprig_get_theme_version() );
}

function wprig_get_theme_version() {
    return '1.0.0';
}
```

One concern that comes to mind is the burden to make sure `wprig_get_theme_version()` gets updated. We can use `wp_get_theme()` inside of `wprig_get_theme_version()` to make sure it always matches whats in style.css but a string would be more performant.

lumpysimon:
Definitely +1 for adding versioning to scripts and style registration. 

What is the benefit to using `filemtime` locally and the theme version number in production? I normally just use `filemtime` on its own, like this:

```
$styles = array(
	'comments',
	'content',
	'sidebar',
	'widgets',
	'front-page'	
);

foreach ( $styles as $style ) {

	$name    = 'wprig-' . $style;
	$file    = 'css/' . $style . '.css';
	$url     = trailingslashit( get_theme_file_uri() ) . $file;
	$path    = trailingslashit( get_template_directory() ) . $file;
	$version = filemtime( $path );

	wp_register_style(
		$name,
		$url,
		array(),
		$version
	);

}
```

mor10:
One tricky thing about version suffixes on JS and CSS files: they should only be updated if the file has actually been updated to take advantage of browser caching. That means over time, different files will have different version numbers. This also has to be respected on full rebuilds.

hellofromtonya:
Let me answer your question, @lumpysimon:

>What is the benefit to using `filemtime` locally and the theme version number in production?

In development, we want the freshest version of our assets. Now once we package up and ship the theme though, the _assets are frozen and will not change_.  Stick with me in this train of thought.

Running `filemtime()` takes a little bit of processing time for _each asset_ as PHP has to check the filesystem for the file's modification timestamp.

But wait a minute, once we ship, the assets do not change.  Therefore, a more performant strategy is to use a set version number instead of incurring the extra processing time to lookup the modification timestamp.

There are a couple of schools of thought on how to set that version number:

1. Use the latest released version number for all of the assets in the package (i.e. theme in this case).
2. Set a version number for each of the assets in the package and only uptick each _when_ that asset changes in a release.

### Option 1

Option 1 is very simple and straight forward from a developer point-of-view.  We use the _theme's release version_.  That version is set in the `style.css` header file as WordPress uses it to notify when to update. 

Using this strategy, we as the developers only have to update the version in one file in the theme's package.  There's less risk of error due to forgetting to update each of the asset versions throughout the theme.

The downside to this strategy is that the browser will request the latest versions of all the assets, even if they have not changed.

Result: Easier for the developer.  Less risk of errors on release.  But a performance hit on the client side on the first visit.

### Option 2

With this strategy uses the theme's released version but then manually sets the version only on the asset file(s) that changed in that particular release.  So if you modified only the stylesheet, then only that asset file gets the new release version, i.e. none of the scripts get it.  If you modified the stylesheet and one of the scripts in a release, then only those 2 files get the new version number.

This approach is more work for the developer, as s/he has to remember to _manually modify_ the enqueue version number for each of the affected assets.  More work for us and more risk of an error.

The benefit though is:  better performance on the client side, as the browser only has to request the assets that changed instead of all the assets.

hellofromtonya:
If I may, @lumpysimon I'd like to offer you a code tip.  Look at the repeating code inside of the loop.  What do you notice?

```
	$url     = trailingslashit( get_theme_file_uri() ) . $file;
	$path    = trailingslashit( get_template_directory() ) . $file;
```

The code to get and prepare the root URI and directory locations is being repeated and every single loop iteration.  Over and over again.  

That's less performant as PHP has to run those functions on every single loop.

Wait a minute.  These root locations do not change in the given code. Right?

A more performant approach is to process and then cache them into a variable _before_ doing the loop.  Then use the variable inside of the loop.

lumpysimon:
@hellofromtonya 

> Running `filemtime()` takes a little bit of processing time for each asset as PHP has to check the filesystem for the file's modification timestamp.

Ah yes of course, I hadn't considered that.

As you say there's not a single straightforward answer, as each approach has pros and cons. I'm not sure what the ideal solution is.

bamadesigner:
Hey gang. Thanks for being patient while I had to take a break for WPCampus. It feels like we're a little stuck here, trying to figure out the best method to ensure assets are updated when needed but not too often. Let me think about this a little. Any new trains of thought?

greatislander:
[Laravel Mix](https://github.com/JeffreyWay/laravel-mix) and the [Sage](https://github.com/roots/sage) build process both write MD5 hashes for each file to a JSON asset manifest. The Laravel Mix manifest can be used to retrieve a unique version for use in in `wp_register_*` or `wp_enqueue_*`; Sage adds the hash directly to the filename and the full URL to the hashed filename is retrieved from the asset manifest, with no version required for `wp_register_*` or `wp_enqueue_*`. The advantage of this approach is that if the compiled file hasn't changed, the "version" remains consistent. If something along these lines sounds like an interesting avenue for further exploration I would be happy to work on a PR.

miklb:
> Any new trains of thought?

I'd vote for Option 1 with a goal of option 2

At least for me with https locally CSS gets cached and browser sync will not work.

mor10:
`v2.0` was just updated (and basically rewritten). Please sync with the upstream branch, test the new code to see if your proposal is still valid, and if so update your PR.

felixarntz:
I actually noticed this is now being taken care of by #176, which is the preferred approach over this. The other PR mitigates the need to define the theme version anywhere in the code entirely, so it's cleaner. If needed, let's continue discussion there.

PR #41, opened 2018-06-29
change wprig_fonts_url logic to filter and loop through array of fonts

davidsword:
## Description

`wprig_fonts_url()` was a little hard to use, manually find+replacing desired fonts, copy/pasting code to add a new font. reworking the function a bit makes it more of a helper function that can be used via a filter within something like `wprig_setup()`, instead of manually editing the function. Example:

```php
function wprig_setup() { 

  // ...

  add_filter('wprig_register_google_font', function( $default_fonts ) {
    // add a font to default ones
    return array_merge($default_fonts, array('Open Sans' => '300,700,700i'));
    // ..or..
    // use only these fonts
    return array(
      'Open Sans'    => '300,400,400i,600,700,700i',
      'Merienda One' => '',
    );
  });

  // ..

}
```

## List of changes

## Checklist:
- [ ] My code is tested. (I did not test out that the 'on'/'off' logic, just assuming it still works)

### DIFF
--- a/dev/functions.php
+++ b/dev/functions.php
@@ -179,39 +179,44 @@ add_filter( 'embed_defaults', 'wprig_embed_dimensions' );
  * Register Google Fonts
  */
 function wprig_fonts_url() {
-	$fonts_url = '';
-
+	$font_families    = array();
+	$fonts_url        = '';
+	$fonts_default    = array(
+		'Roboto Condensed' => '400,400i,700,700i',
+		'Crimson Text'     => '400,400i,600,600i',
+	);
 	/**
-	 * Translator: If Roboto Sans does not support characters in your language, translate this to 'off'.
+	 * Filters default google fonts
+	 *
+	 * use filter in wprig_setup();
+	 *
+	 * @param array $fonts_default array of fonts to use
 	 */
-	$roboto = esc_html_x( 'on', 'Roboto Condensed font: on or off', 'wprig' );
-	/**
-	 * Translator: If Crimson Text does not support characters in your language, translate this to 'off'.
-	 */
-	$crimson_text = esc_html_x( 'on', 'Crimson Text font: on or off', 'wprig' );
+	$reg_fonts = apply_filters( 'wprig_register_google_font', $fonts_default );
 
-	$font_families = array();
+	foreach ( $reg_fonts as $font_name => $font_varients ) {
+		/**
+		 * Translator: If font does not support characters in your language, translate this to 'off'.
+		 */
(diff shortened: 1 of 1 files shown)


mor10:
Related to #27. The idea here is good but I need to see a concrete example of how an end-user would set this up. The goal is to make it as easy and low-touch as possible.

bamadesigner:
I'm always team "add a filter", but I think we can clean up wprig_fonts_url() even more.

Here's some changes I made: https://github.com/wprig/wprig/commit/4275c51a98db67378804a4ea74dabbaada8ca61f

I introduced `wprig_get_google_fonts()`, which takes our default fonts and runs them through the "wprig_google_fonts" filter. Then `wprig_fonts_url()` calls `wprig_get_google_fonts()` to get the fonts and return the Google Fonts URL, if fonts are defined.

I removed `urlencode()` from being run for `add_query_arg()` because its run inside `add_query_arg()`. I also removed the "latin" subset from being defined in the Google Fonts URL because that's set by default for Google Fonts.

Then, I changed `wprig_gutenberg_styles()` and `wprig_styles()` to only load Google Fonts if fonts (and therefore the URL) are defined.

Thoughts?

mor10:
Can someone clean up whatever is causing Travis to get all itchy?

bamadesigner:
Hi @mor10 and @davidsword! I'm back in action after WPCampus. How are we looking on this PR?

mor10:
Need to consolidate this with #27 and the stray `update/fonts` branch somehow. Ideas welcome.

hellofromtonya:
@mor10 Reviewing @bamadesigner's code, I think her code in branch `update/fonts` already takes care of this PR and mine in PR #27.  

- [Fonts filter is found here](https://github.com/wprig/wprig/blob/update/fonts/dev/functions.php#L197)
- Enqueuing optimization to only enqueue when fonts are available is [done here](https://github.com/wprig/wprig/blob/update/fonts/dev/functions.php#L291)

@davidsword take a look at what @bamadesigner [did here to add a fonts filter](https://github.com/wprig/wprig/blob/update/fonts/dev/functions.php#L197).  Does it address your concerns?

```
function wprig_get_google_fonts() {
	$fonts_default = array(
		'Roboto Condensed' => array( '400', '400i', '700', '700i' ),
		'Crimson Text'     => array( '400', '400i', '600', '600i' ),
	);
	/*
	 * Filters default Google fonts.
	 *
	 * @param array $fonts_default array of fonts to use
	 */
	return apply_filters( 'wprig_google_fonts', $fonts_default );
}
```

bamadesigner:
In an attempt to finalize this PR and #27, I've combined into #65.

Issue #35, opened 2018-06-26
Potential gulp improvements


ataylorme:
## Issue Overview

I have some personal gulp preferences from things I learned implementing gulp 4 in other projects. I am not sure if these are welcome improvements to WP Rig or not. 

Please let me know if you would consider some, or all, of these changes. If so I am happy to submit a pull request but don't want to spent time doing without validation of the improvement ideas.

### Monolithic gulp file

[`gulpfile.babel.js`](https://github.com/wprig/wprig/blob/master/gulpfile.babel.js) is a monolithic file that can be modularized

#### Suggested Solution

`gulpfile.babel.js` is written in ES6+ and can make use of imports. Here is [an example](https://github.com/ataylorme/Advanced-WordPress-on-Pantheon/blob/master/web/wp-content/themes/twentyseventeen-child/gulpfile.babel.js)

### Manually loading gulp plugins

[`gulpfile.babel.js`](https://github.com/wprig/wprig/blob/master/gulpfile.babel.js) manually loads gulp plugins. This currently adds 23 lines to the file.

#### Suggested Solution

Use [`gulp-load-plugins`](https://www.npmjs.com/package/gulp-load-plugins) to store gulp plugins in a single variable.

### Misc ES6+ Syntax updates

The `gulpfile.babel.js` uses some older conventions, such as [concatenating strings with `+`](https://github.com/wprig/wprig/blob/master/gulpfile.babel.js#L69).

#### Suggested Solution

Update `gulpfile.babel.js` to consistently use modern conventions. In this specific example using [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

### Use [`pump`](https://www.npmjs.com/package/pump) instead of `pipe`

gulp 4 suggests the use of `pump` instead of `pipe`. See [why use pump](https://github.com/gulpjs/gulp/tree/master/docs/why-use-pump).

#### Suggested Solution

Update `gulpfile.babel.js` to use `pump` instead of `pipe`. Here is [an example](https://github.com/ataylorme/Advanced-WordPress-on-Pantheon/blob/master/web/wp-content/themes/twentyseventeen-child/gulp/buildScript.js).

## Related Issues and/or PRs

Unsure if I should wait for the work in #30 to be merged before starting more gulp work.

hellofromtonya:
Excellent points, @ataylorme.  +1 from me to move forward.

[...]

PR #47, opened 2018-07-06
gulp improvements

ataylorme:
## Description
Addresses issue #35 by adding the following gulp improvements

## List of changes
* Use `gulp-load-plugins` to store gulp plugins in a single variable.
* Use `pump` instead of `pipe`
	- See the gulp 4.0.0 docs on [why use pump](https://github.com/gulpjs/gulp/tree/master/docs/why-use-pump)
* Split functions in gulpfile.babel.js into separate files
	- This allows for `gulpfile.babel.js` to simply be a task/workflow manifest and increases readability vs having one monolithic file
* ES6+ Syntax updates
* Use gulp 4 builtin sourcemaps rather than `gulp-sourcemaps`
* Run gulp from `node_modules` in `package.json` scripts
	- Running the global `gulp` will use the user's local gulp installation which may be a different version than the one we expect.
* Define `rootPath` as `process.env.INIT_CWD` to avoid relative directories parsing from the `gulp` subdirectory
	- See the [npm run script docs](https://docs.npmjs.com/cli/run-script) and [gulp-cli env docs](https://www.npmjs.com/package/gulp-cli4#environment)
* Add options to gulp-string-replace
	- Silence `gulp-string-replace` logging each change
	- Set `gulp-string-replace` search to string from the default of regex as the patterns being used were not regular expressions

## Checklist:
- [x] My code is tested.

**Note:** I tested all gulp functionality manually, there are no automated tests.

### DIFF
--- a/gulp/.eslintrc.json
+++ b/gulp/.eslintrc.json
@@ -1,0 +1,14 @@
+{
+    "parserOptions": {
+        "ecmaVersion": 6,
+        "sourceType": "module"
+    },
+    "env": {
+        "node": true
+    },
+    "extends": "eslint:recommended",
+    "rules": {
+        "semi": 2,
+	"no-unused-vars": 2
+    }
+}
--- a/gulp/browserSync.js
+++ b/gulp/browserSync.js
@@ -1,0 +1,48 @@
+/* eslint-env es6 */
+'use strict';
+
+// External dependencies
+import requireUncached from 'require-uncached';
+import browserSync from 'browser-sync';
+
+// Internal dependencies
+import {paths} from './constants';
+
+/**
+ * Conditionally set up BrowserSync.
+ * Only run BrowserSync if config.browserSync.live = true.
+ */
+
+// Create a BrowserSync instance:
+export const server = browserSync.create();
+
+// Initialize the BrowserSync server conditionally:
+export function serve(done) {
+    // get a fresh copy of the config
+    const config = requireUncached(paths.config.themeConfig);
+
+	if (config.dev.browserSync.live) {
+		server.init({
+			proxy: config.dev.browserSync.proxyURL,
+			port: config.dev.browserSync.bypassPort,
+			liveReload: true
+		});
+	}
+	done();
+}
+
+// Reload the live site:
+export function reload(done) {
+	// get a fresh copy of the config
+    const config = requireUncached(paths.config.themeConfig);
+    
+	if (config.dev.browserSync.live) {
+		if (server.paused) {
+			server.resume();
+		}
+		server.reload();
+	} else {
+		server.pause();
+	}
+	done();
+}
--- a/gulp/bundle.js
+++ b/gulp/bundle.js
@@ -1,0 +1,31 @@
+/* eslint-env es6 */
+'use strict';
+
+// External dependencies
+import {src, dest} from 'gulp';
+import pump from 'pump';
+import requireUncached from 'require-uncached';
+
+// Internal dependencies
+import {paths, gulpPlugins} from './constants';
+
+/**
+ * Create zip archive from generated theme files.
+ */
+export default function bundle(done) {
+    // get a fresh copy of the config
+    const config = requireUncached(paths.config.themeConfig);
+
+	pump([
+        src(paths.export.src),
+        gulpPlugins.if(
+            config.export.compress, 
+            gulpPlugins.zip(`${config.theme.name}.zip`), 
+            dest(`${paths.export.dest}${config.theme.name}`)
+        ),
+        gulpPlugins.if(
+            config.export.compress, 
+            dest(paths.export.dest)
+        ),
+    ], done);
+}
--- a/gulp/constants.js
+++ b/gulp/constants.js
@@ -1,0 +1,93 @@
+/* eslint-env es6 */
+'use strict';
+
+// External dependencies
+import requireUncached from 'require-uncached';
+export const gulpPlugins = require('gulp-load-plugins')();
+
+// gulp string replace options
+export const gulpReplaceOptions = {
+    logs: {
+      enabled: false
+    },
+    searchValue: 'regex',
+};
+
+// Root path is where npm run commands happen
+export const rootPath = process.env.INIT_CWD;
+
+// get a fresh copy of the config
+export const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
+
+// Project paths
+export const paths = {
+	config: {
+		cssVars: `${rootPath}/dev/config/cssVariables.json`,
+		themeConfig: `${rootPath}/dev/config/themeConfig.js`
+	},
+	php: {
+		src: [
+            `${rootPath}/dev/**/*.php`,
+            `!${rootPath}/dev/optional/**/*.*`,
+        ],
+		dest: `${rootPath}/`
+	},
+	styles: {
+		src: [
+            `${rootPath}/dev/**/*.css`,
+            `!${rootPath}/dev/optional/**/*.*`
+        ],
+		dest: `${rootPath}/`,
+		sass: [`${rootPath}/dev/**/*.scss`]
+	},
+	scripts: {
+		src: [
+            `${rootPath}/dev/**/*.js`, 
+            `!${rootPath}/dev/**/*.min.js`, 
+            `!${rootPath}/dev/js/libs/**/*.js`, 
+            `!${rootPath}/dev/optional/**/*.*`, 
+            `!${rootPath}/dev/config/**/*`,
+        ],
+		min: `${rootPath}/dev/**/*.min.js`,
+		dest: `${rootPath}/`,
+		libs: `${rootPath}/dev/js/libs/**/*.js`,
+		libsDest: `${rootPath}/js/libs/`,
+		verboseLibsDest: `${rootPath}/verbose/js/libs/`,
+	},
+	images: {
+		src: [
+            `${rootPath}/dev/**/*.{jpg,JPG,png,svg}`,
+            `!${rootPath}/dev/optional/**/*.*`,
+        ],
+		dest: `${rootPath}/`
+	},
+	languages: {
+		src: [
+            `${rootPath}/**/*.php`,
+            `!${rootPath}/dev/**/*.php`,
+            `!${rootPath}/verbose/**/*.php`,
+        ],
+		dest: `${rootPath}/languages/${config.theme.name}.pot`
+	},
+	verbose: `${rootPath}/verbose/`,
+	export: {
+		src: [
+            `${rootPath}/**/*`,
+            `!${rootPath}/${config.theme.name}`,
+            `!${rootPath}/${config.theme.name}/**/*`,
+            `!${rootPath}/dev/**/*`,
+            `!${rootPath}/node_modules`,
+            `!${rootPath}/node_modules/**/*`,
+            `!${rootPath}/vendor`,
+            `!${rootPath}/vendor/**/*`,
+            `!${rootPath}/.*`,
+            `!${rootPath}/composer.*`,
+            `!${rootPath}/gulpfile.*`,
+            `!${rootPath}/gulp/**/*`,
+            `!${rootPath}/package*.*`,
+            `!${rootPath}/phpcs.*`,
+            `!${rootPath}/*.zip`,
+        ],
+		dest: `${rootPath}/`
+	}
+};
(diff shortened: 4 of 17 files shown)


ataylorme:
> Yes, that's my thinking too. I'll try to get #47 merged into the 1.1 branch tomorrow or Wednesday so we can move forward.

@mor10 if you have any questions let me know. I'm happy to set up a co-working session as well.

gulp/.eslintrc.json
```diff
@@ -0,0 +1,14 @@
+{
```
hellofromtonya: I'm thinking that we should have ESLint at the root of the rig for all of the JS scripts including gulp.  Thoughts?

gulp/browserSync.js
```diff
@@ -0,0 +1,48 @@
...
+ * Only run BrowserSync if config.browserSync.live = true.
+ */
+
+// Create a BrowserSync instance:
+export const server = browserSync.create();
+
+// Initialize the BrowserSync server conditionally:
+export function serve(done) {
+    // get a fresh copy of the config
+    const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Instead of using only the `rootPath`, let's use the `paths.config.themeConfig` here and throughout the scripts.  This strategy let's avoid reuse the paths, i.e. keeping it DRY.

```
const config = requireUncached(paths.config.themeConfig);
```

gulp/browserSync.js
```diff
@@ -0,0 +1,48 @@
...
+			liveReload: true
+		});
+	}
+	done();
+}
+
+// Reload the live site:
+export function reload(done) {
+	// get a fresh copy of the config
+    const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Use the `path.config.themeConfig` here.

gulp/browserSync.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-env es6 */
+'use strict';
+
+// External dependencies
+import requireUncached from 'require-uncached';
+import browserSync from 'browser-sync';
+
+// Internal dependencies
+import {rootPath} from './constants';
```
hellofromtonya: Expand to import `paths`.

gulp/bundle.js
```diff
@@ -0,0 +1,31 @@
...
+
+// Internal dependencies
+import {rootPath, paths, gulpPlugins} from './constants';
+
+/**
+ * Create zip archive from generated theme files.
+ */
+export default function bundle(done) {
+    // get a fresh copy of the config
+    const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Same here using `paths`:

```
const config = requireUncached(paths.config.themeConfig);
```

gulp/constants.js
```diff
@@ -0,0 +1,93 @@
...
+// External dependencies
+import requireUncached from 'require-uncached';
+export const gulpPlugins = require('gulp-load-plugins')();
+
+// gulp string replace options
+export const gulpReplaceOptions = {
+    logs: {
+      enabled: false
+    },
+    searchValue: 'string',
```
hellofromtonya: The `searchValue` option is problematic.  With it set, we are not replacing the text-domains or function prefixes in PHP.  Whoops.  By removing it, everything replaces as it should.

gulp/php.js
```diff
@@ -0,0 +1,69 @@
...
+import colors from 'ansi-colors';
+
+// Internal dependencies
+import {paths, rootPath, gulpPlugins, gulpReplaceOptions} from './constants';
+
+// We are on the first run by default
+let isFirstRun = true;
+
+// Grab the initial config so we can check if the theme slug/name change later
+let initialConfig = require(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Use `paths` here instead of string literal:

```
let initialConfig = require(paths.config.themeConfig);
```

gulp/php.js
```diff
@@ -0,0 +1,69 @@
...
+// Stash theme config
+let themeConfig = initialConfig.theme;
+
+/**
+ * PHP via PHP Code Sniffer.
+ */
+export default function php(done) {
+
+    // get a fresh copy of the config
+    const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Use `paths` here instead of string literal:

```
const config = requireUncached(paths.config.themeConfig);
```

gulp/sassStyles.js
```diff
@@ -0,0 +1,29 @@
...
+
+// Internal dependencies
+import {rootPath, paths, gulpPlugins} from './constants';
+
+/**
+ * Sass, if that's being used.
+ */
+export default function sassStyles(done) {
+    // get a fresh copy of the config
+   const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Use `paths` here instead of string literal:

```
const config = requireUncached(paths.config.themeConfig);
```

gulp/scripts.js
```diff
@@ -0,0 +1,34 @@
...
+
+// Internal dependencies
+import {rootPath, paths, gulpPlugins, gulpReplaceOptions} from './constants';
+
+/**
+ * JavaScript via Babel, ESlint, and uglify.
+ */
+export default function scripts(done) {
+    // Get a fresh copy of the config
+    const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Use `paths` here instead of string literal:

```
const config = requireUncached(paths.config.themeConfig);
```

gulp/scripts.js
```diff
@@ -0,0 +1,34 @@
...
+
+	pump([
```
hellofromtonya: Fix the indent here to align.

gulp/styles.js
```diff
@@ -0,0 +1,58 @@
...
+
+// Internal dependencies
+import {rootPath, paths, gulpPlugins, gulpReplaceOptions} from './constants';
+
+/**
+* CSS via PostCSS + CSSNext (includes Autoprefixer by default).
+*/
+export default function styles(done) {
+   // get a fresh copy of the config
+   const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Use `paths` here instead of string literal:

```
const config = requireUncached(paths.config.themeConfig);
```

gulp/translate.js
```diff
@@ -0,0 +1,30 @@
...
+
+// Internal dependencies
+import {rootPath, paths, gulpPlugins} from './constants';
+
+/**
+ * Generate translation files.
+ */
+export default function translate(done) {
+    // Get a fresh copy of the config
+    const config = requireUncached(`${rootPath}/dev/config/themeConfig.js`);
```
hellofromtonya: Use `paths` here instead of string literal:

```
const config = requireUncached(paths.config.themeConfig);
```

gulp/.eslintrc.json
ataylorme: @hellofromtonya the issue I see with that is the `env:node`, which is true for gulp but not the JS being written for the front end scripts. Without that ESLint throws warnings for using `process` and other Node variables in gulp.

We could do inline ESLint ignore comments in the gulp files to get around that or maybe make creative use of the [ESLint configuration](https://eslint.org/docs/user-guide/configuring) file precedence.

hellofromtonya: Yup, you're right @ataylorme.  Let's do this to keep it clean.  Leave it as you have it.  Then we can take on a new issue for how we want to proceed with adding ESLint to the root of the rig.  Sound good?

ataylorme: Works for me. There was no ESLint config in WP Rig but I wanted to have one to help me write better code when working with gulp for this PR

gulp/php.js
```diff
@@ -0,0 +1,69 @@
...
+
+	// We should rebuild if this is the first run OR the theme slug/name have changed
+	let isRebuild = isFirstRun ||
+		( themeConfig.slug !== config.theme.slug ) ||
+        ( themeConfig.name !== config.theme.name );
+
+	if ( isRebuild ) {
+		themeConfig.slug = config.theme.slug;
+        themeConfig.name = config.theme.name;
+        log(colors.yellow(`Rebuilding ${colors.bold('all')} the PHP files, this may take a while...`));
```
hellofromtonya: Excellent job @ataylorme adding the log messages here.  Very nice 👍

gulp/styles.js
```diff
@@ -0,0 +1,58 @@
...
+           })
+       ]),
+       gulpPlugins.stringReplace('wprig', config.theme.slug, gulpReplaceOptions),
+       gulpPlugins.stringReplace('WP Rig', config.theme.name, gulpReplaceOptions),
+       dest(paths.verbose),
+       gulpPlugins.if(
+           !config.dev.debug.styles, 
+           gulpPlugins.cssnano()
+        ),
+       dest(paths.styles.dest, {sourcemaps: true}),
```
hellofromtonya: Really dig the sourcemaps.  Excellent job @ataylorme! 👍

gulp/constants.js
ataylorme: d'oh! I updated it to explicitly be regex rather than deleting the value in case the default changes in the future.

gulp/styles.js
ataylorme: Thanks! I love that gulp 4 has them built-in

gulp/php.js
ataylorme: Thanks for the +1 @hellofromtonya. It actually isn't too slow for me with the base PHP files but I can imagine it slowing down as more files are added during theme development.

ataylorme:
> here's the PHP code review part. I did test it on my machine. Running tests on CSS now.

Thanks for the review @hellofromtonya, especially the feedback on DRY with the config path. I totally missed that one.

gulp/php.js
hellofromtonya: It isn't slow for me either.  But the indication in the console helps tell a story of what's going on.  I like that.

gulp/.eslintrc.json
mor10: The ESLint config sits in `package.json` https://github.com/wprig/wprig/blob/8bfd9cbe4f7055df2ea7d519666c77803bab1528/package.json#L62

gulp/sassStyles.js
```diff
@@ -0,0 +1,29 @@
...
+
+// Internal dependencies
+import {paths, gulpPlugins} from './constants';
+
+/**
+ * Sass, if that's being used.
+ */
+export default function sassStyles(done) {
+    // get a fresh copy of the config
+   const config = requireUncached(paths.config.themeConfig);
```
ataylorme: @hellofromtonya do you think using `requireUncached` is necessary? It might be cleaner to read the config once in constants, export it and then import it here rather than break cache to re-read.

I'm not sure if we expect the debug config to be toggled while gulp is running. Reading uncached each time, like is happening now, will allow that to happen.

hellofromtonya: I'd say yes, @ataylorme.  You can be working along and then discover: "Whoops, I misspelled the prefix or theme name."  Right? That can happen.  In that case, you want the ability to change the theme configuration on-the-fly with gulp running, and have it rebuild for you.

ataylorme: Makes sense, I just wanted to confirm.

mor10:
I've run all my tests and am ready to move forward with this PR @ataylorme. Can you please [squash your commits](https://github.com/wprig/wprig/wiki/How-to-squash-commits) so I can merge this PR cleanly with the 1.1 branch?

mor10:
Also, before squashing, can you take out the copying of php files to the `./verbose/` folder? It causes some unexpected behavior. See #51 for details.

ataylorme:
@mor10 updated for #51 and squashed. Let me know if you need anything else.

mor10:
Merged to 1.1. Can someone (@ataylorme / @bamadesigner / @hellofromtonya) check to make sure everything works after merge?

joseph-farruggio:
I'm the author of [issue #51](https://github.com/wprig/wprig/issues/51) mentioned by @mor10 . After switching to the v1.1 branch i'm still seeing verbose templates appear in template select dropdown. Note: I only see verbose templates when viewing an edit view with Gutenberg. Classic edit views do not show any verbose templates.

Issue #48, opened 2018-07-06
Notice: add_theme_support() was called incorrectly.


roncham:
Setting colors using multiple parameters is deprecated. Please pass a single parameter with an array of colors. See https://wordpress.org/gutenberg/handbook/extensibility/theme-support/ for details.

In dev/functions.php around line 104.

```
Notice: add_theme_support() was called <strong>incorrectly</strong>. Setting colors using multiple parameters is deprecated. Please pass a single parameter with an array of colors. See https://wordpress.org/gutenberg/handbook/extensibility/theme-support/ for details. Please see <a href="https://codex.wordpress.org/Debugging_in_WordPress">Debugging in WordPress</a> for more information. (This message was added in version 3.4.0.) in C:\wamp64\www\ptplayground\wp-includes\functions.php on line 4161
Call Stack
#	Time	Memory	Function	Location
1	0.0005	428904	{main}( )	...\post.php:0
2	0.2829	27623112	apply_filters( )	...\post.php:155
3	0.2829	27623512	WP_Hook->apply_filters( )	...\plugin.php:203
4	0.2829	27624904	gutenberg_init( )	...\class-wp-hook.php:286
5	0.2837	27659896	require_once( 'C:\wamp64\www\ptplayground\wp-admin\admin-header.php' )	...\gutenberg.php:191
6	0.2848	27669272	do_action( )	...\admin-header.php:97
7	0.2848	27669648	WP_Hook->do_action( )	...\plugin.php:453
8	0.2848	27669680	WP_Hook->apply_filters( )	...\class-wp-hook.php:310
9	0.2959	27723200	gutenberg_editor_scripts_and_styles( )	...\class-wp-hook.php:286
10	0.3764	29145880	_doing_it_wrong( )	...\client-assets.php:1157
11	0.3764	29146888	trigger_error ( )	...\functions.php:4161
```

Also editor-color-palette arrays are missing the slug eg:
```
array(
			'name'  => 'Dusty orange',
			'slug'  => 'dusty-orange',
			'color' => '#ED8F5B',
		),
```

webmandesign:
@roncham To fix the issue please try to apply the [code changes](https://github.com/wprig/wprig/pull/49/files) from pull request #49

roncham:
Thanks. I found the solution right before you pushed the changes.

webmandesign:
I'd say we should leave this open until guys merge my pull request.

mor10:
Pull request was just merged. I'm closing this issue.

PR #49, opened 2018-07-06
Theme support for Gutenberg color palette with a single array attribute

webmandesign:
## Description

Updating Gutenberg editor color palette theme support: colors are now passed using a single array attribute instead of multiple attributes for each color separately.

@see  https://github.com/WordPress/gutenberg/issues/6425

Addresses issue #48 

## List of changes

* Adding an array wrapper around existing palette colors declarations.
* Making color name translatable.
* Adding color slug.
* Changing color hex value to lower case.

## Checklist:

- [ ] My code is tested.

### DIFF
--- a/dev/functions.php
+++ b/dev/functions.php
@@ -103,47 +103,47 @@ function wprig_setup() {
 	 */
 	add_theme_support( 'editor-color-palette', array(
 		array(
-			'name'  => esc_html__( 'Dusty orange', 'wprig' ),
+			'name'  => __( 'Dusty orange', 'wprig' ),
 			'slug'  => 'dusty-orange',
 			'color' => '#ed8f5b',
 		),
 		array(
-			'name'  => esc_html__( 'Dusty red', 'wprig' ),
+			'name'  => __( 'Dusty red', 'wprig' ),
 			'slug'  => 'dusty-red',
 			'color' => '#e36d60',
 		),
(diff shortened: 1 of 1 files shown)


spencerfinnell:
A slug is required now too: https://github.com/WordPress/gutenberg/blob/master/docs/extensibility/theme-support.md#block-color-palettes -- would be good to add with this.

webmandesign:
@spencerfinnell Yes, you are right. Thanks for pointing this out! I've modified the code even further.

dev/functions.php
```diff
@@ -101,44 +101,53 @@ function wprig_setup() {
 	 *
 	 * @link https://wordpress.org/gutenberg/handbook/extensibility/theme-support/
 	 */
-	add_theme_support( 'editor-color-palette',
+	add_theme_support( 'editor-color-palette', array(
 		array(
-			'name'  => 'Dusty orange',
-			'color' => '#ED8F5B',
+			'name'  => esc_html__( 'Dusty orange', 'wprig' ),
```
felixarntz: Using `esc_html__()` is unnecessary here (and in the other similar occurrences), escaping should always happen right before outputting which is not what we're doing here. Simply using `__()` is more appropriate.

webmandesign: @felixarntz Done ;)

Issue #50, opened 2018-07-10
Make the theme actually require PHP 7.0 with graceful fallback


felixarntz:
## Issue Overview
As stated in the project readme, the theme requires PHP 7.0. We should be enforcing that so that we can securely use also PHP-related modern practices, and do this gracefully though so that users are notified about the requirement being insufficient instead of encountering a WSOD.
The WordPress theme repository will soon support a new header too in which the minimum required PHP version can be specified (this is already the case for plugins and for themes it has been agreed on as well, just needs to be implemented).

## Possible Solution
* We should move all actual functions out of `functions.php` into subfiles in the `inc` directory that will then be included. This will make the "bootstrap" file plain and simple and in the process also likely improve file organization (most of the code could probably go into a file `inc/setup.php`).
* By having `functions.php` only contain inclusions of other files, we can now theoretically use PHP 7.0 code in all other included files. Only `functions.php` must remain parseable by PHP 5.2.
* What we need to do then is implement a `inc/back-compat.php` file which also needs to be parseable by PHP 5.2. That file should add the necessary hooks for when the PHP version is insufficient, notifying the user of that requirement and stopping related functionality in a graceful way.
* That file should then be included by `functions.php` _if_ the PHP version is lower than 7.0. If it is, after including that file we should simply put a return statement, in order to prevent the other (regular) files from being included.

The above is already an established practice for ensuring the minimum WordPress version requirement is met, used for example by Twenty Seventeen. I also have it implemented in [my own boilerplate](https://github.com/felixarntz/theme-boilerplate/blob/master/inc/back-compat.php), and it would be easy to adapt it to work for the PHP version instead of the WordPress version.

mor10:
I love this idea, and it falls in line with what WP Rig already does with modules, just in a much more aggressive way. Can you put together a PR where you take this as far as you think it should go? Considering we're rewriting the entire Gulp function for 1.1, we might as well rewrite all of functions.php while we're at it.

felixarntz:
I'll work on it tomorrow!

PR #59, opened 2018-07-17
Ensure that the theme requirements are met, otherwise bail and warn the user

felixarntz:
## Description
Addresses issue #50

This PR ensures that the theme requirements are checked, and if not met, the theme isn't initialized and the user receives a warning about this instead. The techniques implemented here have been already established by popular themes such as Twenty Seventeen.

The minimum required WP version is 4.5 (as used in `phpcs.xml.dist`).
The minimum required PHP version is 7.0 (as mentioned in `README.md`).

## List of changes

* All functions have been moved out of `functions.php`. The functions that were previously part of it have been moved to separate files `inc/setup.php` and `inc/assets.php`. This allows to use modern PHP code in those files.
* Constants have been introduced for the minimum required WP and PHP versions.
* In order to catch these constants with the gulp replacing mechanism, support for that has been added. Since the value of the constant prefix can easily be determined automatically from `theme.slug` (and make it uppercase), it is not mandatory to provide that argument, but a developer still can if they prefer to use something else. A utility function `getThemeConfig()` takes care of that handling.
* In `functions.php` initially it is checked whether the two minimum requirements are met. If not, a new `inc/back-compat.php` file is loaded, and then the process bails, not initializing the theme any further.
* `inc/back-compat.php` contains logic to warn the user about the issue and keep the user from running into fatal errors possibly caused by the incompatibility.
* Going forward, we need to ensure that `functions.php` and `inc/back-compat.php` must remain parseable by PHP 5.2. The two files should generally not be edited, unless to add a new require statement to the `functions.php`. All other PHP files can use PHP 7.0 code as necessary.
* The `.travis.yml` and `phpcs.xml.dist` files have been adjusted to use PHP 7.0 as minimum version to test against.

## Checklist:
- [x] My code is tested.


mor10:
This looks useful, and would fit nicely into the full upgrade with v1.1. However, that would mean rewriting most of the Gulp features. Can you work with @ataylorme to get this PR aligned with the new Gulp process in 1.1?

ataylorme:
@mor10 should the PR be made against `v1.1` rather than `master`? If so I am happy to help convert it

ataylorme:
Went ahead and did the conversion for `v1.1`. It was easy enough to pull in @felixarntz's branch and clean up merge conflicts. @felixarntz can you review/test #69? Things looked good for me but I would like a +1

felixarntz:
Closing in favor of #69 which is the more up-to-date follow-up for this one.

Issue #74, opened 2018-07-29
Notice: is_amp_endpoint was called incorrectly


iliman:
## Issue Overview
When the AMP plugin and the lazy-loading feature are activated, the following notice is issued:

```
Notice: is_amp_endpoint was called incorrectly. is_amp_endpoint() was called before the 'parse_query' hook was called. This function will always return 'false' before the 'parse_query' hook is called. Please see Debugging in WordPress for more information. (This message was added in version 0.4.2.) in /www/wp.test/wp-includes/functions.php on line 4161
```

## Describe your environment
WP Rig 1.0.2
WordPress 4.9.7
Apache/2.4.33 (Unix)
PHP/7.2.4

## Steps to Reproduce

1. Set the `define('WP_DEBUG', true);` in wp-config.php;
2. Activate the `AMP for WordPress` plugin;
3. Activate the lazy-loading feature;
4. Visit any AMP page.

## Possible Solution
The function `wprig_lazyload_images()`, which calls the `wprig_is_amp()` function inside, must only be called after the `parse_query` action.

PR #75, opened 2018-07-29
Fix is_amp_endpoint was called incorrectly

iliman:
## Description

Addresses issue #74

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/dev/pluggable/lazyload/lazyload.php
+++ b/dev/pluggable/lazyload/lazyload.php
@@ -42,6 +42,7 @@ function wprig_lazyload_images() {
 	add_filter( 'wp_kses_allowed_html', 'wprig_allow_lazy_attributes' );
 
 }
+add_action( 'wp', 'wprig_lazyload_images' );
 
 /**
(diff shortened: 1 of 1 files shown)


felixarntz:
Good catch! I'm wondering whether `wp` is the best hook to attach this too. At what point do we actually need the logic to be available? I think it should be initialized as late as possible, in order to not be unnecessary overhead for requests where it might not even be applicable (REST API etc.).

iliman:
@felixarntz considering that the `wprig_lazyload_images()` function contains some hooks inside that must be executed on time(`wp_enqueue_scripts` is the earliest), we have the following corridor:

image

It turns out that `get_header` is the latest hook we can use. What do you think?

felixarntz:
Thanks for the quick response @iliman, and for looking into the topic more closely! In that case, I think the `wp` hook is a solid choice. Let's move forward with this.

bamadesigner:
Awesome! I'm going to merge into master since this is throwing notices. Thanks @iliman and @felixarntz.

Issue #43, opened 2018-07-01
Use a theme slug that is more than one word


felixarntz:
## Issue Overview

We need to consider whether the theme `slug` in `themeConfig.js` will always be the only parameter that needs to be covered when making initial naming replacements. The slug at the moment is only a single word, however what happens if the the slug would be something consisting of two words? If I want to have `my-theme`, this would only work as a textdomain replacement, but no longer as function name prefix replacement (we'd probably need to sanitize it to `my_theme` there. If the code at some point includes classes or namespaces, we would also need to account for `My_Theme`.

## Possible Solution

I think the starter theme would be better off using distinct `wp_rig` and `wp-rig` strings, possibly even `WP_Rig`. That would make replacements in such cases easily possible while at the moment it uses the simplest use-case.

## Related Issues and/or PRs

This came up during #31.

mor10:
Flagging this as a bug to give it priority for 1.1

mor10:
Pick a day you want to do this on and we'll do a PR freeze until the PR is submitted, tested, and merged.

felixarntz:
I'll implement this next Tuesday.

felixarntz:
Fixed via #93

PR #93, opened 2018-08-28
Fix theme name and slug replacement process

felixarntz:
## Description

Addresses issue #43 

This PR aims at fixing numerous issues with the string replacements, when setting up the theme name and slug:
* Multiple variants of the slug need to be replaced: hypen-delimited (for textdomain), underscore-case, underscore-case capitalized (for constants), camel-case (for JavaScript) and camel-case with the first letter in lowercase (for JavaScript variables)
* In order to make these replacements work, the default theme slug needs to consist of a minimum of two terms, so "wprig" doesn't work. "wp-rig" does though and is a good alternative.
* Replacements must also occur in JavaScript and CSS files, not only in PHP files.

We might also consider introducing a replacement for the theme homepage, so that it is easy to replace "https://github.com/wprig/wprig/" since that occurs in code as well.

## List of changes
* [x] All relevant occurrences of "wprig" in `/dev` files have been replaced with "wp-rig", "wp_rig", "WpRig" or "wpRig" respectively. Excluded are those strings that are unrelated to the code string replacement, for example "wprig" occurring in URLs.
* [x] All relevant occurrences of "WPRIG" in `/dev` files have been replaced with "WP_RIG".
* [x] Gulp now replaces all different variants in all PHP files properly.
* [ ] Gulp now replaces all different variants in all JavaScript files properly.
* [ ] Gulp now replaces all different variants in all CSS files properly.

## Checklist:
- [ ] My code is tested.

### DIFF
--- a/gulp/php.js
+++ b/gulp/php.js
@@ -8,17 +8,9 @@ import log from 'fancy-log';
 import colors from 'ansi-colors';
 
 // Internal dependencies
-import {paths, rootPath, gulpPlugins, gulpReplaceOptions} from './constants';
+import {paths, rootPath, gulpPlugins, gulpReplaceOptions, nameFieldDefaults} from './constants';
 import {getThemeConfig} from './utils';
 
-const nameFieldDefaults = {
-	slug          : 'wp-rig',
-	name          : 'WP Rig',
-	underscoreCase: 'wp_rig',
-	constant      : 'WP_RIG',
-	camelCase     : 'WpRig',
-	camelCaseVar  : 'wpRig',
-};
 const nameFields = Object.keys( nameFieldDefaults );
 
 // We are on the first run by default
@@ -41,8 +33,9 @@ export default function php(done) {
 	// We should rebuild if this is the first run OR the theme name fields have changed
 	let isRebuild = isFirstRun;
 	let i = 0;
-	while ( ! isRebuild || i >= nameFields.length ) {
+	while ( ! isRebuild && i < nameFields.length ) {
 		isRebuild = isRebuild || themeConfig[ nameFields[ i ] ] !== config.theme[ nameFields[ i ] ];
+		i++;
 	}
 
 	if ( isRebuild ) {
--- a/gulp/php.js
+++ b/gulp/php.js
@@ -11,6 +11,16 @@ import colors from 'ansi-colors';
 import {paths, rootPath, gulpPlugins, gulpReplaceOptions} from './constants';
 import {getThemeConfig} from './utils';
 
+const nameFieldDefaults = {
+	slug          : 'wp-rig',
+	name          : 'WP Rig',
+	underscoreCase: 'wp_rig',
+	constant      : 'WP_RIG',
+	camelCase     : 'WpRig',
+	camelCaseVar  : 'wpRig',
+};
+const nameFields = Object.keys( nameFieldDefaults );
+
 // We are on the first run by default
 let isFirstRun = true;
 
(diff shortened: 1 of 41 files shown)


felixarntz:
@ataylorme I started work on this. Most of the code changes (and the part which will unfortunately introduce merge conflicts over time) are the replacements of all the occurrences of those strings in the codebase. Don't worry too much about those.

The actual relevant parts are the changes in the `/gulp` directory, those would need eyes. Also I'd like to discuss how we should tackle making those replacements in the JavaScript and CSS files. Should we outsource the replacement logic into a separate .js file and then use that in the respective files for the JS and CSS workflows? Or should we put all of it into the `php.js` file (which then would need to be renamed to something like `replaceStrings.js`)?

mor10:
Ok. I'm putting a moratorium on all PR merges until this one is done.

gulp/php.js
```diff
@@ -25,47 +35,52 @@ let themeConfig = initialConfig.theme;
...
 
-	// We should rebuild if this is the first run OR the theme slug/name have changed
-	let isRebuild = isFirstRun ||
-		( themeConfig.slug !== config.theme.slug ) ||
-        ( themeConfig.name !== config.theme.name ) ||
-        ( themeConfig.constant !== config.theme.constant );
+	// We should rebuild if this is the first run OR the theme name fields have changed
+	let isRebuild = isFirstRun;
+	let i = 0;
+	while ( ! isRebuild || i >= nameFields.length ) {
```
ataylorme: `i` is never incremented. Is it needed?

gulp/php.js
```diff
@@ -25,47 +35,52 @@ let themeConfig = initialConfig.theme;
...
+			bin: `${rootPath}/vendor/bin/phpcs`,
+			standard: 'WordPress',
+			warningSeverity: 0
+		}),
+		// Log all problems that was found
+		gulpPlugins.phpcs.reporter('log'),
+	];
+
+	Object.keys( nameFieldDefaults ).forEach( nameField => {
+		steps.push( gulpPlugins.stringReplace( nameFieldDefaults[ nameField ], config.theme[ nameField ], gulpReplaceOptions ) );
```
ataylorme: Nice use of a loop here to feed `steps` into `pump`

muggeridge:
Is WP Rig ready for theme production now?

gulp/php.js
felixarntz: Oops, I missed that.

felixarntz:
@ataylorme Thanks for the feedback, I addressed those issues.

Regarding testing, we should implement the replacements for JS and CSS files before we start testing the whole thing. Can you share your thoughts on my question above? Once we have decided on an approach, I can take care of implementing.

ataylorme:
> Should we outsource the replacement logic into a separate .js file and then use that in the respective files for the JS and CSS workflows? Or should we put all of it into the `php.js` file (which then would need to be renamed to something like replaceStrings.js)?

I am not in favor of doing it all in `gulp/php.js` as PHP code sniffing is also done there, so it would be misleading to name the file `replaceStrings.js` and still have other functions going on.

`gulp/scripts.js` and `gulp/styles.js` also do string replacement. 

Since `pump` takes an array we could export a `stringReplacement` array with the replacement functionality from `gulp/utils.js` and use it in `gulp/php.js`, `gulp/scripts.js` and `gulp/styles.js`. 

As some items in the `pump` process of those files occur after the replacement we could have a `beforeReplacement` and `afterReplacement` array and [`concat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) those with the `stringReplacement` array imported from `gulp/utils.js` in the middle.

felixarntz:
@ataylorme Agreed. Will move forward with that approach.

felixarntz:
@ataylorme I implemented the new utility function and it's now used for PHP, JS and CSS. Review-ready!

ataylorme:
@felixarntz the code looks great! I’m on holiday until Tuesday but will test the functionality when I’m back at the computer next week.

ataylorme:
@felixarntz I tested this with the default `WP Rig` as well as `WP Rig Custom Name!` and it is working well in both cases. I think it is odd though the user must provide `slug` and `name` in `dev/themeConfig.js`.

What do you think about only specifying `name` and generating `slug` automatically? We could use a method similar to [sanitize_title_with_dashes](https://codex.wordpress.org/Function_Reference/sanitize_title_with_dashes).

At a minimum we should add a comment to `dev/themeConfig.js` specifying the format for `slug`.

mor10:
`what_if_my_theme_name_is_very_long_and_contains_superfluously_elaborate_and_lomgwinded_words`

felixarntz:
The latest commit adds a comment about the required slug format and auto-generates it from the name if not present. I think we should still keep the `slug` in the config array as it indicates that it can be different from the name.

ataylorme:
@mor10 do you want to do any testing? The code looks good to me and all my testing went well.

mor10:
Works in my tests. I'm merging. Any issues that arise after merge can be fixed as they appear.

Issue #22, opened 2018-06-17
Add integration tests for functionality


felixarntz:
## Issue Overview
While themes generally should not contain too much functionality (plugin territory), they always _do_ contain some functionality (and they have to), for example to define the theme behavior, to register sidebars, customizer integration etc.

Just like plugins, this functionality should be tested. As a first step, I suggest adding an infrastructure for integration tests, which work relatively similar to how WordPress plugins are commonly tested: WordPress is loaded in a specific version, and then the theme's individual functions are checked for whether they interact with WordPress in the way they should.

Actual unit tests would also be useful, however that should be part of a separate issue. Integration tests are already widely known and used for WordPress plugins, so we can do this as a first iteration.

## Possible Solution
I have recently set this up for a theme I have developed. I use a similar technique like the typical plugin tests (see https://make.wordpress.org/cli/handbook/plugin-unit-tests/):
* PHPUnit is used to initialize the test suite and run tests.
* The WordPress core test suite is loaded (detected via environment variable, or a fallback assuming that the theme is actually in the themes directory of the tested WordPress instance).
* Per a custom theme tests bootstrap script, WordPress is tricked into thinking that our theme is the currently active theme.
* We can now run tests with the theme being active, using WordPress core functionality.
* Integration tests should be run in CI as well. I suggest we introduce a new `.travis.yml` environment variable `INTEGRATION`, and if that is set (via `INTEGRATION=1`), PHPUnit is installed, the WordPress core development version is downloaded, and our tests are run. The WordPress version should be dynamically set via another `WP_VERSION` environment variable.
* In terms of the test matrix, I think two new setups should be used: Both on the latest PHP version, both with `INTEGRATION=1`. One should then use the minimum required WordPress version of the theme, the other the newest version the theme has been tested with.

_Aside:_ Before working on a PR for this, I'll wait until the existing related PRs have been closed, in order to not run into too many merge conflicts later.

dingo-d:
Or just add/setup tests using [codeception](https://codeception.com/for/wordpress)?

felixarntz:
I think as a first iteration it is already a great step to use the core test suite to test the theme. We can then consider further testing implementations later, I'd suggest. Using the core test suite is also less problematic in terms of finding a framework to use - as that one is certainly maintained.

dingo-d:
[...]

PR #114, opened 2018-09-17
Add PHP unit and integration tests infrastructure

felixarntz:
## Description

Addresses issue #22 

This PR introduces a modern PHP tests infrastructure to the the theme, for both unit and integration tests.

## List of changes
* Use PHPUnit for both unit and integration testing.
* For unit tests (that require to be isolated from actually running WordPress), use the `Brain\Monkey` library that is established as a unit testing tool for WordPress.
* For integration tests, use the WordPress core test suite. This kind of setup is typically used by WordPress plugins, however they typically end up mixing unit and integration tests with this approach (with true unit testing hardly being possible though).
* Three specific test classes (one for unit tests, two for integration tests) have been implemented as a start. Providing more solid test-coverage through additional tests can happen through follow-up issues.
* As a guideline for the future:
    * Unit tests should only test a single function / method in isolation.
    * Integration tests should test whether WordPress core is aware of the theme functionality as expected.
* Unit tests and integration tests can be individually run, and separate configuration files are provided for them. For easy access, just call `composer run-script phpunit-dev` or `composer run-script phpunit-integration-dev` respectively.
* Tests are integrated with Travis-CI:
    * Unit tests are run for the newest and oldest supported PHP version.
    * Integration tests are always run for the newest supported PHP version, and against the newest and oldest supported WordPress version.
    * Two additional `.travis.yml` jobs test against the nightly PHP and WordPress version respectively. Those are allowed to fail, but will still help in recognizing future errors early.
* I added myself as codeowner for the tests part, as I'm happy to focus on issues and PRs for this area in the future.

## Checklist:
- [x] My code is tested.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,9 @@
   "require-dev": {
     "wp-coding-standards/wpcs": "^1.0.0",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
-    "wimg/php-compatibility": "^8.1"
+    "wimg/php-compatibility": "^8.1",
+    "phpunit/phpunit": "^6",
+    "brain/monkey": "^2"
   },
   "scripts": {
     "install-codestandards": [
@@ -15,6 +17,8 @@
     "phpcs-dev": "\"vendor/bin/phpcs\"",
     "run-phpcs": [
       "@phpcs-dev"
-    ]
+    ],
+    "phpunit-dev": "\"vendor/bin/phpunit\"",
+    "phpunit-integration-dev": "\"vendor/bin/phpunit\" -c phpunit.integration.xml.dist"
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,6 @@
       "@phpcs-dev"
     ],
     "phpunit-dev": "\"vendor/bin/phpunit\"",
-    "phpunit-integration-dev": "\"vendor/bin/phpunit\" -c phpunit.integration.xml.dist"
+    "phpunit-integration-dev": "\"vendor/bin/phpunit\" -c \"phpunit.integration.xml.dist\""
   }
 }
--- a/composer.json
+++ b/composer.json
@@ -4,6 +4,7 @@
   "type": "wordpress-theme",
   "license": "GPL-2.0",
   "require-dev": {
+    "php": ">=7.0",
     "wp-coding-standards/wpcs": "^1.0.0",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
     "wimg/php-compatibility": "^8.1",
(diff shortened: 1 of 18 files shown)


mor10:
Re-running Travis after merging other requests.

mor10:
I'm merging this. We should add some documentation in `README.md` and the wiki about how to use these tests.

felixarntz:
@mor10 Feel free to open an issue for this and assign me!

composer.json
```diff
@@ -15,6 +23,8 @@
     "phpcs-dev": "\"vendor/bin/phpcs\"",
     "run-phpcs": [
       "@phpcs-dev"
-    ]
+    ],
+    "phpunit-dev": "\"vendor/bin/phpunit\"",
+    "phpunit-integration-dev": "\"vendor/bin/phpunit\" -c \"phpunit.integration.xml.dist\""
```
dingo-d: Maybe add additional scripts to skip creating code coverage report (if it's generated in the first place).

Or just add to the readme that people can pass additional arguments in the composer script using 

```bash
composer phpunit-dev -- --no-coverage
```

And `--filter` for running single test suite.

Also adding `@php` will automatically resolve to whatever php process is currently being used.

https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts

dev/tests/phpunit/bootstrap-common.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * WP Rig common tests bootstrap script.
+ *
+ * @package wp_rig
+ */
+
+// Disable xdebug backtrace.
+if ( function_exists( 'xdebug_disable' ) ) {
+	xdebug_disable();
```
dingo-d: This will completely disable xdebug? How do you generate code coverage?

I'd disable xdebug in the travis, so that it's faster there...

dingo-d:
One thing that can be added to the `require-dev` in the composer is the [`codedungeon/phpunit-result-printer`](https://github.com/mikeerickson/phpunit-pretty-result-printer), which will make the output of the testing a bit prettier :)

Also a `printerClass` should be added in the `phpunit.xml.dist` and `phpunit.integration.dist.xml` file for it to work

```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit
  ...
  printerClass="Codedungeon\PHPUnitPrettyResultPrinter\Printer"
  ...
```

composer.json
dingo-d: Just took a look at the xml file and there is no `<logging>` part. Should this be added?

felixarntz:
@dingo-d Thanks for the feedback! Regarding coverage, let's set this up in a separate PR. Feel free to create one to start working on this. We might want to integrate with a service such as CodeClimate or similar for this.

And about the testing output, I'm not familiar with the library you're suggesting, but it surely is worth investigating. Please open an issue / PR for that as well. :)

Issue #104, opened 2018-09-10
WPRIG (uppercase) is not renamed


benoitchantre:
## Issue Overview
If you add a constant, like `WPRIG_THEME_VERSION`, `WPRIG` is not replaced by an uppercase version of the theme's slug. 

## Expected Behavior
`WPRIG` should be replaced by the theme slug in uppercase.

## Current Behavior
`WPRIG` is not replaced.

ataylorme:
@benoitchantre can you try the `v1.1` branch? THe process was overhauled in #93

benoitchantre:
@ataylorme I did a merge on a work in progress project:

- PHP files look good (except errors from PHP CS, probably from my code)
- in navigation.js, I get the following error: `mythemeslugScreenReaderText is not defined`

felixarntz:
@ataylorme I figured out why this happens: In `gulp/scripts.js`, the tasks only run if the relevant files have been edited since (via `gulpPlugins.newer`). This needs to be changed to also take into account whether the theme configuration has changed (see all the custom code currently part of `gulp/php.js`). Preferably, this should be added to `gulp/styles.js` as well - it doesn't matter at this point, but if someone added CSS classes prefixed with `wp-rig-` for example, they would run into the same issue.

@benoitchantre Until this is actually fixed, you should be able to workaround the issue by simply making a random change in the JavaScript file and saving, so that the time it was last updated is changed.

ataylorme:
Thanks for identifying @felixarntz. `gulp-newer` allows you to check files outside the stream and, if they are newer than the destination files, to rebuild all files in the stream. I opened #123 to take advantage of this. A review when you can would be much appreciated.

benoitchantre:
Thank you for the tip @felixarntz.

mor10:
#123 is merged, so closing. Feel free to open if it still doesn't work.

PR #123, opened 2018-09-19
Watch theme config for changes and rebuild all the things

ataylorme:
## Description
Watch theme config for changes and rebuild all the things.

## List of changes
* Uses the [`gulp-newer` `ext` option](https://www.npmjs.com/package/gulp-newer#neweroptions) to allow files to be rebuilt when the source hasn't changed but the theme config has
    * Runs all default tasks, except the BrowserSync server, when the theme config has changed
            * Also alerts the user about the theme config change and rebuild
    * Addresses #104

## Checklist:
- [x] My code is tested.

### DIFF
--- a/gulp/jsLibs.js
+++ b/gulp/jsLibs.js
@@ -13,9 +13,12 @@ import {paths, gulpPlugins} from './constants';
  */
 export default function jsLibs(done) {
 	pump([
-        src(paths.scripts.libs),
-        gulpPlugins.newer(paths.scripts.verboseLibsDest),
-        dest(paths.scripts.verboseLibsDest),
-        dest(paths.scripts.libsDest),
-    ], done);
+		src(paths.scripts.libs),
+		gulpPlugins.newer({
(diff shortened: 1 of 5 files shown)


felixarntz:
@ataylorme Looks great to me! Can you add this handling to the stylesheets to (e.g. `styles.js`)?

ataylorme:
> Can you add this handling to the stylesheets to (e.g. styles.js)?

@felixarntz I am not sure what you mean by this.

`gulp-newer` is not used in `gulp/styles.js`. The updated watch task for theme configuration runs the `styles` task if the configuration changes.

felixarntz:
@ataylorme Ahh, my bad, I must have misread the code there.

ataylorme:
Cool - anything else you noticed? I tested and think this is good to go but would like at least a +1 before we call it good to merge

mor10:
Works on my end.

ataylorme:
Great, thanks for testing @mor10. Anything else before it gets merged? You'll have to do that as I don't have permission

PR #132, opened 2018-09-27
Merge master into v2.0 branch

felixarntz:
Currently, the `v2.0` branch is a few commits behind `master`. That shouldn't be the case, this PR simply merges the not yet included changes from `master` into `v2.0`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ WP Rig requires the following dependencies. Full installation instructions are p
 
 - [PHP](http://php.net/) 7.0
 - [npm](https://www.npmjs.com/)
-- [Composer](https://getcomposer.org/)
+- [Composer](https://getcomposer.org/) (installed globally)
 
 ### How to install WP Rig:
 1. Clone or download this repository to the themes folder of a WordPress site on your development environment.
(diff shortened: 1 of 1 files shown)


felixarntz:
@mor10 Something must have went wrong there, as `v2.0` is still the 5 commits behind `master`.

mor10:
I'll take a closer look.

mor10:
Weird. I think they might have been dislocated because I used "Squash and merge" rather than just "Merge". Anyway, it looks to me like all the updates are in the `v2.0` branch so it should be OK, at least from a pure code perspective. Not entirely sure how (and whether there's any need) to "fix" this if the only issue is a disconnect within Git, not actual code missing. Thoughts @felixarntz?

felixarntz:
@mor10 It is just confusing. I agree there's no code-wise benefit, but I think we should fix it - I guess by simply merging instead of squashing and merging. Obviously low priority, but in the future people might be wondering why there is something in v1 that is not in v2.

mor10:
Pull down the latest version of v2.0. I did some updates and it looks to me like the changes were baked in. Master has also diverged from v2.0 so the problem is no longer "real" if that makes any sense.

Issue #129, opened 2018-09-24
Idea: restructure WP Rig to support a source/build style


ataylorme:
## Issue Overview
The current approach of editing files in a nested `dev` directory within the production version of the theme surfaces some issues.

* Deleted files are not removed from the theme (#101)
* Files not included in production are included (#118)
* It is not intuitive to create new WordPress PHP files in the `dev` directory

## Expected Behavior
When working with other software that has to be built for production there is typically a source and a build directory without the source directory and files being included in/nested in the build directory.

## Possible Solution
Rather than having a `dev` directory nested in the WP Rig theme I propose WP Rig, as hosted on GitHub, be a development theme that is never actually activated. The `dev` directory is removed and all source files, task processes/automation, etc. are in the top level WP Rig development theme.

Gulp is then used to generate a production version of the theme. This would exclude non-proudction assets, automation scripts/tooling, tests, etc. The production version of the theme is what would be activated.

When developing all work is done on the development version of the theme which will update the activated production theme. The existing `bundle` command can also be used to generate a `.zip` of the production ready theme.

This keeps the source/development files and production files separated and well organized, more closely matching a traditional source/build model.

mor10:
Some background on why the folder structure is the way it is right now:

From a user standpoint, two of the main goals for WP Rig were:
a) for it to work within a development environment out of the box meaning once you added WPRig to the `/wp-content/themes/` folder and ran the process, a theme would emerge.
b) the theme development folder (currently `/dev/`) would be the only place most developers made changes to code, and that this folder look as close to a "normal" theme as possible.

These two principles are core to the entire project and should not be strayed from. As long as your proposal does not collide with them, I am interested in seeing where this idea leads.

To be quite honest, I'm not entirely sure what you're proposing here, so an example of some sort would be useful ;o)

ataylorme:
I'm proposing not having a nested `dev` directory in the active theme. Instead have `wp-content/themes/wp-rig` be the place to develop and `wp-content/themes/my-theme` be the artifact that WP Rig produces. 

Items not necessary in production, such as `composer.json`, would be excluded from the production theme. Conversely production only-items, like minified CSS, would be excluded from the development theme.

A lot of what is in `dev` now, the files developers work on, would be in the development theme.

I think another benefit to this approach is a developer could create a theme with WP Rig and just hand off the production theme to a client and the client would not need WP Rig knowledge to use the built theme.

ataylorme:
[...]

PR #133, opened 2018-10-06
Restructure WP Rig for v2.0

ataylorme:
# Work In Progress
Do **NOT** merge until all the boxes below are checked.

## Description
Please see #129 for the discussion of changes. These are **breaking** changes intended for a major release.

## List of changes
**File Restructuring**
- [x] Move all files and directories in `dev` up one level to the root directory
    - [x] Exception: `dev/css` should become `/css/src`
    - [x] Exception: `dev/js` should become `/js/src`
    - [x] Exception: `dev/images` should become `/images/src`
- [x] use the [`theme_scandir_exclusions` filter](https://core.trac.wordpress.org/browser/tags/4.9/src/wp-includes/class-wp-theme.php#L1155) to ignore the `/optional` directory

**The Development gulp Process (`npm run dev`) Does The Following:**
- [x] Compiles unminified CSS from `/assets/css/src/` to `/assets/css/` passing through PostCSS and StyleLint
    - [x] Add `.min` to minified file names
- [x] Compiles unminified JS from `/assets/js/src/` to `/assets/js/` passing through Babel and ESLint
    - [x] Add `.min` to minified file names
- [x] Optimizes images from `/assets/images/src/` to `/assets/images/` passing through [`gulp-imagemin`](https://www.npmjs.com/package/gulp-imagemin)
- [x] Checks PHP files against PHPCS
- [x] Watches for changes to PHP files, JS, CSS and images then and re-runs their corresponding build task
- [x] Initializes BrowserSync
- [x] Hot reloads/refreshes BrowserSync whenever PHP, JS, CSS or image files change

**The Production gulp Process (`npm run bundle`) Does The Following:**
- [x] Compiles unminified CSS from `/css/src/` to `/css/` passing through PostCSS and StyleLint
    - [x] Add `.min` to minified file names
- [x] Compiles unminified JS from `/js/src/` to `/js/` passing through Babel and ESLint
    - [x] Add `.min` to minified file names
- [x] Checks PHP files against PHPCS
- [x] Copy files from the dev theme directory to a production theme directory
    ~- [ ] Only copy files based on a whitelist defined in `/config/themeConfig.js`.~
    - [x] The whitelist defaults to: all PHP files (excluding the `optional` directory), the `css` and `js`directories (excluding their `src` subdirectories),  the `images` directory, the `/pluggable` directory, `readme.txt` and `screenshot.png`
- [x] Runs prefix renaming process on all PHP and JS files in the production theme directory
- [x] Runs image optimization on the `/assets/images/` folder in the production theme directory
- [x] Runs translate processes on the production theme directory
- [x] Creates a `.zip` archive of the production theme directory
    - [x] Can be disabled with a setting in `/config/themeConfig.json`
~- [ ] The production build directory can be disabled with a setting in `/config/themeConfig.json`. In this case, a temporary directory will be used to create the `.zip` then deleted after the production build process~

**String Replacement Task**
~- [ ] New gulp task to run the string replacement one-time on the theme files (PHP, CSS, JS)~
    - Now #165 
- [x] All name variations configurable in theme config
- [x] String replacement is skipped if the name configuration is left as the default

## Checklist:
- [x] My code is tested.


ataylorme:
I see 3 options regarding the debugging settings for CSS/JS in `themeConfig.js`. I am leaning towards option 1 but would like additional opinions.

1) Remove the settings as we have source maps so debugging is still possible with modern browser dev tools, which will show the source
1) Don't minify the CSS/JS at all during development, only in the production build.
1) Have the debug settings in `themeConfig.js` just toggle minification of CSS/JS. When `debug` is `true` for either CSS or JS then they don't get minified, just copied as-is from `src` to the main CSS/JS directories (but are still checked for coding standards).

ataylorme:
Okay I _think_ the reworked gulp build process (dev mode/non-production) is done.

Note one significant change: I added `filemtime` as the version to JS/CSS files. Without this browser cache caused changes to not be seen without a manual hard refresh, negating the effects of an automated BrowserSync reload.

**Please help test**. I would like the default gulp dev task/watch process to be thoroughly tested and approved before I move on to creating the new gulp production build process.

mor10:
an idea: create a `style.src.css` file in root and have that file be processed into `style.css` by the task. I'm experimenting to see if I can wrangle your code into doing that.

ataylorme:
>  One immediate thing: style.css in the root is not processed by the build task. That breaks everything as the file has a bunch of variables which need to be replaced. Not entirely sure how to handle this as style.css needs to remain in root and can't be renamed to style.min.css

WordPress requires `style.css` for the theme name, etc. but we don't have to enqueue it. What if it was there with just the WordPress required comment. We could then take the rest of the CSS and put it into `css/src/global.css` or similar so that global styles would be processed and enqueue the resulting CSS file.

mor10:
That's a radical departure from the norm which opens the door to all sorts of possibilities and does away with my ridiculous idea. So yes, that makes a lot of sense in my head. Thoughts @bamadesigner?

mor10:
Thinking of how we're going to rethink all of CSS for `v2.1` and explore BEM/ITCSS/modules etc, keeping all CSS in the `/css/` folder and conforming to WordPress requirements only to get things to work lays a solid platform imo.

mor10:
and yes, `global.css` is a good name for that file.

ataylorme:
@mor10 actually I want to modify my proposal to just move everything below the comment in `style.css` to a global CSS file and still keep `style.css` enqueued. This would solve the processing/variable concern but still allow users to add CSS to `style.css` if they wanted to.

mor10:
`style.css` needs to stay in root for everything to work, but I don't think we need to enqueue it at all. I'd rather have developers get used to doing all their CSS work in the `/css/` folder and leave `style.css` to only do the job of talking to WordPress. If they start adding CSS to `style.css`, we'll be back in "how to process that file" territory.

mor10:
btw I got this working by rejigging some files. Do you want my code? And if so, how do I get that to you since I'm working off a fork off your branch?

ataylorme:
> I'd rather have developers get used to doing all their CSS work in the /css/ folder and leave style.css to only do the job of talking to WordPress. If they start adding CSS to style.css, we'll be back in "how to process that file" territory.

Works for me.

> btw I got this working by rejigging some files. Do you want my code? And if so, how do I get that to you since I'm working off a fork off your branch?

I allowed editors to update the PR but it is a small change so I don't mind making it

ataylorme:
@mor10 the styles from `/styles.css` have been moved into `/css/global.min.css`. All processed CSS and JS files now have `.min` in them as well.

`inc/assets.php` was only registering styles, not enqueueing them. I added a new function to enqueue styles as well.

After your feedback about preferring to list CSS files rather than glob CSS files I also added a `wp_rig_get_css_files` function that returns a list of CSS files, filterable by `wp_rig_css_files`. The functions to register and enqueue styles both fetch the list from this new function.

The styles are registered and enqueued separately so that they can be enqueued to the footer, rather than the `<head>`. The `preload` functionality still exists.

Let me know if I missed anything with registering/enqueueing the styles.

mor10:
The stylesheets should not be enqueued. They are progressively loaded in-body at component level in various template files.

ataylorme:
Okay, I will look at it again tomorrow. I think the global styles we removed from style.css need to be enqueued but I will remove the others.

mor10:
Yes, the global styles are global and should be enqueued. The rest are modular.

ataylorme:
> Yes, the global styles are global and should be enqueued. The rest are modular.

I updated the `inc/assets.php` so that only the global styles are enqueued.

mor10:
Amazing work @ataylorme. I'm running more tests today. Take some time off from this monster project to be with your family

ataylorme:
> Take some time off from this monster project to be with your family

I plan on it! Tomorrow will consist of having friends/family over, firing up the grill and watching football.

kevinwhoffman:
Hi everyone, I'm just getting caught up on all things WP Rig and reviewing the build process and directory structures. One thing that stood out to me is that the images `src` and `dist` directories appear to the be same:

https://github.com/wprig/wprig/blob/e5ad2fb69222da28ed041bff91f82dd41d1aa8ff/gulp/constants.js#L60-L63

It might be intentional, but it has the side effect of creating irreversible changes regarding image quality since any images committed to version control will likely have been minified and overwritten by gulp prior to committing. This could also cause image quality to further degrade each time the minification process occurs as the quality gets worse with each iteration.

The alternative, and my preference based on past projects, is to set separate `src` and `dist` directories so minification changes can happen without degrading the quality of the source image over time.

ataylorme:
Thank you for the input @kevinwhoffman 

> It might be intentional, but it has the side effect of creating irreversible changes regarding image quality since any images committed to version control will likely have been minified and overwritten by gulp prior to committing. This could also cause image quality to further degrade each time the minification process occurs as the quality gets worse with each iteration.

Yes, the current process will overwrite any images with the optimized version. The image task uses [gulp-newer](https://www.npmjs.com/package/gulp-newer) so images shouldn't be saved over more than once.

Can you elaborate on _why_ keeping the original, non-optimized images is important to you?

> The alternative, and my preference based on past projects, is to set separate src and dist directories so minification changes can happen without degrading the quality of the source image over time.

Yes, having a `/images/src` directory is one alternative. Another alternative would be to only process images when doing the production build. One of the goals of this restructuring is to make the gulp development process as fast as possible and not having to watch/optimize images at all in development would certainly speed things up.

kevinwhoffman:
> Can you elaborate on why keeping the original, non-optimized images is important to you?

For the same reason it's important to preserve the original image when it's uploaded to WordPress. If you decide to change minification settings at any point in the future, you want to generate the newly minified image from the original source and not from a compressed version so as to avoid compound compression.

ataylorme:
> If you decide to change minification settings at any point in the future, you want to generate the newly minified image from the original source and not from a compressed version so as to avoid compound compression.

It sounds like excluding image processing during the dev process will solve this. The next time a production build is run all images will be processed and any minification setting changes will be applied.

mor10:
You beat me to it. Image optimization can be a very slow process so it should only happen in the bundle task when the theme is prepped for shipping. There could be  a stand-alone task dfor just compressing images as well perhaps?

kevinwhoffman:
I agree image optimization is slow, but my main concern was whether minified images were being committed to Git. I'm not totally clear on the output of the production build. If the production tasks generates a zip that is output somewhere outside of version control, then that would satisfy the concern of committing compressed images. But if the production task instead optimizes the images and leaves them in the same directory, then it would seem those compressed images would enter the Git staging area.

Other assets like scripts and styles have `dist` directories that are kept out of version control and I assume only get processed when a change occurs. I guess what I'm proposing is the same for images.

ataylorme:
> I'm not totally clear on the output of the production build. If the production tasks generates a zip

I haven't built the production build step yet but essentially, yes, it will create a production-ready version of the theme outside version control. This will be a `.zip` bundle and/or (depending on configuration) another top-level theme directory.

The idea being you work with non-optimized assets during development. When development is done and a new version of the theme is ready for release then the production build step can run, either manually or automated (e.g. as part of a CI process), and deployed.

For me personally an ideal workflow would be using the dev person of the theme locally and CI hooked to version control for building + deploying the built version (staging environment on a PR, production on master)

hellofromtonya:
@ataylorme What if we could provide the means for devs to not a build task when the images are already optimized?  What might that look like?  I like the workflow you've proposed.  But @kevinwhoffman is right too as there will devs and edge cases where they don't want to run the optimization build task. 

So the question then is:  How can the rig serve both needs?

I'd imagine that we could add an option to exclude a build task.  What do you think @ataylorme?

hellofromtonya:
>There will always need to be some sort of build task, even for development, to process JS/CSS due to lack of browser support for modern JS and the use of CSS variables.

Sorry for being unclear here, @ataylorme.  I was specifically talking about providing an option to exclude the images from being optimized.  You are right that there will always been the need to run build tasks.

ataylorme:
> I was specifically talking about providing an option to exclude the images from being optimized

I haven't developed the new production bundle task yet so am open to suggestions there. A setting in `/config/themeConfig.js` seems appropriate.

Should we make any other steps, such as JS/CSS minification, optional as well when creating the production bundle?

hellofromtonya:
>I haven't developed the new production bundle task yet so am open to suggestions there. A setting in /config/themeConfig.js seems appropriate.

I like that.  It's clearly a configuration parameter.  +1

kevinwhoffman:
Could we serve the optimized images from `assets/images/dist` and ignore that directory so images are treated just like compiled scripts and styles?

This way we can still use gulp-newer to limit the frequency of image optimization. Then image optimization can happen at any time (in a production build or a one-off task) while ensuring that source images are version controlled and compressed images are only generated as part of the build.

ataylorme:
> Could we serve the optimized images from assets/images/dist and ignore that directory so images are treated just like compiled scripts and styles

I don't have an opinion one way or the other. I just want to build what is most useful. Lets do a hybrid approach and have `/assets/images/src` that compiles to `/assets/images` (no additional `dist` to keep with the same convention we are using for CSS/JS) *and* have a configuration option, which defaults to `true`, that includes optimized images in the production bundle. When explicitly set to `false` the source images will be copied instead of the optimized ones.

kevinwhoffman:
> Lets do a hybrid approach and have /assets/images/src that compiles to /assets/images (no additional dist to keep with the same convention we are using for CSS/JS) and have a configuration option...

That makes sense to keep the directory structure the same as the other assets. Thanks.

felixarntz:
@ataylorme

> Another alternative would be to only process images when doing the production build. One of the goals of this restructuring is to make the gulp development process as fast as possible and not having to watch/optimize images at all in development would certainly speed things up.

I'd prefer if we only built images in production. I don't think for our development unoptimized images are a problem. By only building images in production, we don't need an extra directory, but still maintain original images (in the development version it will never be changed). Regarding the general concern by @kevinwhoffman, I agree that, regardless of what we decide, original images shouldn't be overwritten.

I'm okay with the proposed solution of having images in `assets/images/src` be optimized (or just copied) to `assets/images`, although I'd still say we don't need image optimization during development (as it would slow down the process for little benefit), so we might as well keep images in `assets/images` and only build them for production (where everything goes to a separate directory anyway). The purist in me approves of having `assets/images` and `assets/images/src` because it is similar to how CSS and JS work.

---

I fixed a couple issues with the PHPUnit and PHPCodeSniffer integrations, so that they should now work with the new paths. Another observation, we should probably add adjusting the `README.md` file to the bottom of the task list for this PR. Once the code is ready, that documentation needs to be updated accordingly.

I haven't yet tested in detail, but before I was going to do that, I would like to know @ataylorme, is this already good for testing? When I tried running `gulp`, I ran into an error, and before wasting time diving into this I wanted to make sure.

ataylorme:
> is this already good for testing? When I tried running gulp, I ran into an error, and before wasting time diving into this I wanted to make sure.

The default gulp development task should be okay for testing. You will probably need to get a fresh set of dependencies from npm as things have changed. This alias is my friend `alias npmPlease='rm -rf node_modules/ && rm -f package-lock.json && npm install'`

I'm not sure if this is the case but you should also be running `npm run build` rather than gulp directly.

mor10:
I just run `npm install` any time I pull one of these branches. Seems to work.

ataylorme:
> have `/assets/images/src` compile to `/assets/images` (no additional `dist` to keep with the same convention we are using for CSS/JS)

This has been implemented. I am headed to WordCamp Philadelphia this weekend but would like to start working on the production step when I return to keep this moving forward toward a merge.

Has anyone had a chance to test the dev process/default gulp task? I would like to have approval on the dev process before starting to write the production bundle task.

ataylorme:
> I'd still say we don't need image optimization during development (as it would slow down the process for little benefit), so we might as well keep images in assets/images and only build them for production

In my testing, adding a new image to `/assets/images/src` didn't add very much overhead to the process (138ms). The browser reload needs to occur either way when images are changed so the additional time to optimize isn't really detectable.

> The purist in me approves of having assets/images and assets/images/src because it is similar to how CSS and JS work.

I think having a consistent directory structure for all assets (CSS, JS, and images) is the way to go rather than having images be different so I went with that approach.

felixarntz:
I tested this a little more thoroughly today.

✅ The `npm run build` part worked as expected.

✅ The `npm run translate` part worked as expected. I wonder whether we even need this task though.

✅ The `npm run bundle` part worked as expected.

❌ When I ran `npm run dev`, the assets were built as expected and my browser opened my dev site with the theme. However, when I changed a PHP file and saved it, the watch process would trigger endlessly. I only changed the file and saved it once, but I got the following output:
```
[13:36:57] Starting 'php'...
[13:37:06] Finished 'php' after 8.48 s
[13:37:06] Starting 'reload'...
[13:37:06] Finished 'reload' after 2.9 ms
[13:37:06] Starting 'php'...
[Browsersync] Reloading Browsers...
[13:37:14] Finished 'php' after 8.5 s
[13:37:14] Starting 'reload'...
[13:37:14] Finished 'reload' after 2.3 ms
[13:37:14] Starting 'php'...
[Browsersync] Reloading Browsers...
[13:37:23] Finished 'php' after 8.64 s
[13:37:23] Starting 'reload'...
[13:37:23] Finished 'reload' after 2.03 ms
[13:37:23] Starting 'php'...
[Browsersync] Reloading Browsers...
[13:37:32] Finished 'php' after 8.49 s
[13:37:32] Starting 'reload'...
[13:37:32] Finished 'reload' after 2.37 ms
[13:37:32] Starting 'php'...
[Browsersync] Reloading Browsers...
```

This essentially continued on and on for multiple times. So there seem to be two issues that need to be fixed here:
* When a file is changed, the necessary tasks to process this change should only fire once.
* When a PHP file is changed, nothing should happen, since they don't need any processing. The only thing affected would be the `.pot` translation file, however that doesn't play any role during "live development", so I think it rightly is not part of the `npm run dev` task.

On subsequent tries of changing a PHP file and saving it, I noticed that the aforementioned loop appears to happen not endlessly, but for a couple times (the number of times differs). Most of the times, it went on for quite a bit before I cancelled. One of the times though, it only looped twice - I did the same thing in all cases, so I'm not sure where this comes from.

---
Another small detail I noticed is that on every task I run, I get this initially:
```
[13:53:05] Failed to load external module @babel/register
[13:53:05] Requiring external module babel-register
```
While this doesn't cause the tasks to fail, it seems that this module should be included so that this message doesn't show.

ataylorme:
> When a PHP file is changed, nothing should happen, since they don't need any processing.

@felixarntz PHP files should be code sniffed so we can't ignore them from the development watch tasks.

> However, when I changed a PHP file and saved it, the watch process would trigger endlessly

I'll investigate this, thanks for testing!

ataylorme:
> However, when I changed a PHP file and saved it, the watch process would trigger endlessly

@felixarntz this should be fixed now.

felixarntz:
@ataylorme 
Your fix works for me, all good now.

> PHP files should be code sniffed so we can't ignore them from the development watch tasks.
Ah right. I think it makes sense to continue code sniffing on changes. However, it could probably use some performance improvements. I assume at the moment all PHP files are checked? It would speed it up a lot if only the changed files were tested.

ataylorme:
> I assume at the moment all PHP files are checked? It would speed it up a lot if only the changed files were tested.

Yes, you are correct that all PHP files are code sniffed. Unfortunately, I don't think we can only code sniff changed files, at least not using [`gulp-newer`](https://www.npmjs.com/package/gulp-newer), since the source and destination files are the same. That plugin works with timestamps. If you save a PHP file the timestamp after you save it is used. That timestamp will never be different than the destination as it is the same file. We can detect changed files for CSS and JS since the source and destination files are different.

package.json
```diff
@@ -80,10 +82,11 @@
...
-    "build": "gulp",
+    "build": "gulp buildDev",
+    "dev": "gulp",
     "translate": "gulp translate",
-    "bundle": "gulp bundleTheme",
-    "test": "gulp testTheme",
-    "generateCert": "gulp generateCert"
+    "bundle": "NODE_ENV=production gulp bundleTheme",
+    "generateCert": "gulp generateCert",
+    "devStringReplace": "gulp sourceStringReplacement"
```
felixarntz: I don't like having this task here. It's not of value for the intended workflow, and more importantly can cause problems with that workflow. Since it furthermore adds extra overhead to this PR, I'd prefer to remove it. Sure, someone might wanna approach it this way, but that's not the intention, and it could always be implemented in the respective projects specifically.

_If_ I'm in the minority with saying that it should be removed, I'd at least prefer to only have this task present at the `gulp` level, but not among the other "encouraged" tasks in `package.json`.

package.json
```diff
@@ -22,7 +22,7 @@
     "babel-preset-env": "^1.7.0",
     "babel-preset-es2015": "^6.24.1",
     "babel-register": "^6.26.0",
```
felixarntz: For some reason, I still get the following every time I run a task:
```
[11:25:22] Failed to load external module @babel/register
[11:25:22] Requiring external module babel-register
```

It doesn't cause the task to break, but it seems wrong. Would it be possible to fix it so that notice doesn't show up on every task?

package.json
ataylorme: I'll let @mor10 chime in here since I think he was the one who requested this be possible.

> If I'm in the minority with saying that it should be removed, I'd at least prefer to only have this task present at the gulp level, but not among the other "encouraged" tasks in package.json.

The items in `package.json` aren't just there to be the "encouraged" items. Running gulp with [npm run script]()https://docs.npmjs.com/cli/run-script) allows for the gulp binary installed with npm to be used, rather than requiring the user to install gulp globally.

That being said you can pass extra arguments to items run with npm run script. I added a generic gulp script to npm scripts so that `npm run gulp` can be used. Then, any gulp task can be invoked. For example, `npm run gulp -- images` would run the image task.

Now we can remove `devStringReplace` but keep the gulp task if needed.

package.json
ataylorme: Based on [gulp issue 1631](https://github.com/gulpjs/gulp/issues/1631) I don't think this is an issue.

package.json
ataylorme: @felixarntz Unless you object, let's handle the decision to keep or remove this task in a follow-up issue/PR. It would be a small change so I don't think we should block merging this PR.

package.json
```diff
@@ -80,10 +82,12 @@
   },
   "scripts": {
     "install": "composer install",
-    "build": "gulp",
+    "build": "gulp buildDev",
+    "dev": "gulp",
+    "gulp": "gulp",
```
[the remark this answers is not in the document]
ataylorme: I added `gulp` so it can be used to run other tasks besides the default task. For example, `npm run gulp -- images` would run the image task in isolation. 

While `npm run dev -- images` would do the same thing as they both map to `gulp` I think `gulp` is more intuitive.

If we are going to remove one I would advocate for removing `dev` rather than `gulp` and update the documentation so that `npm run gulp` is the command we recommend for development.

ataylorme:
>  don't like the idea of merging something we might remove. I'd like to hear at least initial thoughts from @mor10 about this before proceeding.

> I do consider whatever scripts are in package.json encouraged. Advanced commands could still be accessible via gulp directly, and the benefit of not having to install gulp globally is great especially for people who are relatively new to this. I'd argue they probably wouldn't wanna diverge from the suggested path anyway though - and having this task in package.json makes it more likely to be used, probably without knowing the consequences.

@felixarntz I removed the source string replacement task entirely. I would rather get this PR merged then discuss in a follow-up issue. If we decide to add it back based on Morten's feedback then we can do so in a new PR.

README.md
```diff
@@ -44,60 +51,72 @@ To take full advantage of the features in WP Rig, your code editor needs support
...
+JavaScript source files are transpiled with [Babel](https://babeljs.io/) and CSS source files are transpiled with [PostCSS](https://postcss.org/) so that you may use modern JavaScript and CSS without worrying about browser support.
 
-JavaScript files are automatically linted using [ESLint](https://eslint.org/) in accordance with [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
+Additionally, JavaScript files are automatically linted using [ESLint](https://eslint.org/) in accordance with [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
 
 PHP and CSS files are automatically linted using [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) in accordance with [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/). To take full advantage of this setup, configure your code editor / IDE to automatically test for the WordPress Coding Standards.
```
mor10: Need to add something about the use of stylelint in the CSS task here. Possibly break PHP and CSS into separate sections to make that clear. Off the top of my head I'm also wondering whether using both PHPCS _and_ stylelint to lint CSS is overkill. Would stylelint suffice?

mor10:
@ataylorme When I run `npm run build`, the process executes down through php and styles, then finishes `buildDev` and exits. BrowserSync is not booted up, and the dev process exits meaning no watching takes place. What am I missing here?

screen shot 2018-11-13 at 3 41 18 pm

ataylorme:
Try npm run gulp. Build is one time processing so that is working as intended

mor10:
Riiight. I see now for `buildDev` it says "Build theme for development without BrowserSync or watching."  When I run `npm run gulp` I get watching and BrowserSync. So is that the plan forward? Or am I misunderstanding?

ataylorme:
That’s the plan. I’m open to feedback if you think the task names or npm aliases should be different

mor10:
I think `npm run dev` should be the main development run, including BrowserSync and watch, but before I make up my mind I want to know why you think it shouldn't be.

mor10:
When I test, the CSS custom properties are not transformed into regular CSS. It must be related to how `postcssPresetEnv` is set up. By default, it doesn't do anything to custom properties, but when `preserve: false` is set, it should transform the values into regular CSS. https://www.npmjs.com/package/postcss-preset-env#preserve

I remember I had some issue with this previously, and figured out a way around it, but I can't recall what that solution was right now.

Out of the box, the custom properties should turn into pure CSS in the minified files. There's a custom property at the very top of `content.min.css` for testing.

ataylorme:
> I think npm run dev should be the main development run, including BrowserSync and watch, but before I make up my mind I want to know why you think it shouldn't be.

I can add that as an alias but I want to keep `npm run gulp` as aside from the default task it allows you to run any arbitrary gulp task through npm. 

For example, `npm run gulp -- images` for the image task. The advantage of running through npm is it uses the gulp binary in `node_modules` and does not require users to install gulp globally.

We could do that with `npm run dev` but `npm run dev -- images` doesn't seem as intuitive for arbitrary gulp tasks,

We could have both `npm run dev` and `npm run gulp` but @felixarntz indicated he was against having two npm scripts mapped to `gulp`. Personally I don't care either way. Can the two of you come to a consensus?

mor10:
I understand and agree with the thinking around arbitrary gulp tasks the way you describe. I also don't think there is any problem in keeping both `npm run gulp` and `npm run build` as aliases for the same thing as long as we only document `npm run build` for this purpose and document `npm run gulp --thing` for only sub-task runs.

That way, if you want to do a development run with all the bells and whistles, you go the `npm run build` route. If you want to do something specific, you go the `npm run gulp --thing` route. Clear separation of concerns, and semantically self-explanatory processes.

mor10:
To the previous comment, I'm tentatively OK with swapping the development command to `npm run dev` as long as there's clear documentation on the difference between `build` and `dev`.

ataylorme:
> I also don't think there is any problem in keeping both npm run gulp and npm run build as aliases for the same thing as long as we only document npm run build for this purpose and document npm run gulp --thing for only sub-task runs.

Both have been added and the README is updated

> To the previous comment, I'm tentatively OK with swapping the development command to npm run dev as long as there's clear documentation on the difference between build and dev.

The README reflects this

README.md
felixarntz: +1 on using only stylelint on CSS.

felixarntz:
+1 on having `npm run dev` as the default development task, as that is pretty much the convention I feel.

I'm still unclear what the benefit of having `npm run gulp` in addition is. I get that it indicates that you can run any Gulp command with it. However, running anything else than the basic commands in the workflow should be considered advanced usage. I think only those commands should be in `package.json`. If you want to run `gulp` directly with certain subtasks, you can still run `gulp` directly. And actually, if you are already familiar with that anyway, you will figure out that `npm run dev` actually just calls `gulp`, so you'll be able to pass arguments to it. I just don't like having `npm run gulp`, which only exists for running advanced tasks, as a command in `package.json`. Sure, we'll document that it is advanced and more specific, but I think as a best practice we should only encourage the commands that actually matter for the intended workflow.

mor10:
@felixarntz has a point: `gulp -- whatever` does the same thing as `npm run gulp -- whatever`, so why do we need the custom setup for `npm run gulp` in `package.json` in the first place? Can't we just as easily say "if you want to run individual processes, use `gulp -- [processName]'? That does away with the entire problem unless I'm missing something. @ataylorme is there a particular reason for defining `gulp` as a separate script in `package.json`? Does it open the door to some functionality I'm not aware of?

ataylorme:
> is there a particular reason for defining gulp as a separate script inpackage.json`? Does it open the door to some functionality I'm not aware of?

Yes, see [my comment above](https://github.com/wprig/wprig/pull/133#discussion_r232389030). Running gulp through npm ensures that the gulp binary installed with npm is used and that it is added to `PATH`.

Without this users would need to install gulp globally. That is an unnecessary step as we are installing gulp with npm. Additionally, a global version could be any version where the version defined in npm is a predictable version defined in `package.json`.

I think having tasks we encourage, `npm run dev`, `npm run build` and `npm run bundle`, as well as the generic `npm run gulp` alias for other gulp tasks covers all use cases.

> `gulp -- whatever does the same thing as npm run gulp -- whatever`

I think you might be misunderstanding the double dash `--`. It is needed in `npm run gulp -- whatever` so npm can determine what arguments should be passed to the script it is calling. gulp, in this case. So `npm run gulp -- whatever` would be the same as `gulp whatever` if you were using a global gulp binary rather than the gulp binary installed with npm.

felixarntz:
I agree with @mor10. I understand that `npm run gulp` gets around installation of Gulp globally, but you could as well use `npm run dev` because it is exactly that, and by looking at the code you'll figure it out. Or you can run `gulp` directly - if you wanna make use of those advanced use-cases, you'll probably have Gulp installed anyway.

mor10:
What happens if we ping @ai and @jonathantneal on this `postcssPresetEnv` issue?

jonathantneal:
Hi there. So, a few things...

Custom Properties are only transformed if required by the supported browsers in your Browserslist file.

Custom Property references, `var()` can only be transformed if we have accsss to the original custom property declaration. If it’s not in the same file being processed, you can use the `importFrom` to reference any external JS, JSON, or CSS file with them. If you’re doing this from PostCSS Preset Env, be sure you’re running the latest version, of check the changelog for any fixes that came after the version you are using that might have affected this.

Are these conditions met?

mor10:
Thanks for piping in @jonathantneal. The code in question is here:

https://github.com/wprig/wprig/blob/b1be54b1d96a378bfa6fc06bc31414f76be8eb4f/gulp/styles.js#L39-L52

In the project the variables (properties and media) are defined in `config/cssVariables.json` and pulled into the process. It is possible this has to do with the BrowserList config like you suggested. What confuses me is this same setup used to work in a previous version of the app.

mor10:
@ataylorme I tested changes to `browserlist`, and that doesn't fix the variable problem. `browserlist` changes does impact other parts of the compiling, so that's not where the problem is. The next question is whether `config/themeConfig.js` is available inside the function or not. 

`importFrom`, mentioned by @jonathantneal, might be an answer? https://www.npmjs.com/package/postcss-preset-env#importfrom

mor10:
Might find some relevant info here: https://github.com/csstools/postcss-preset-env/issues/32
That's where the original code came from when I switched from `cssnext` to `postcss-preset-env`

jonathantneal:
@mor10, changing each of those `variables` arguments with `importFrom: { customProperties: cssVars.variables }`, and `importFrom: { customMedia: cssVars.queries }` should work for now.

They can all live in the same file if they each live in different property names.

mor10:
Reading some of the documentation I'm wondering if we need to add in https://github.com/postcss/postcss-custom-media and https://github.com/postcss/postcss-custom-properties to get this to work...

jonathantneal:
Those are all bundled with PostCSS Preset Env.

ataylorme:
@jonathantneal thank you for the information. I was able to get this working after reading the docs on `importFrom` in [`postcss-custom-media](https://github.com/postcss/postcss-custom-media#importfrom) and [`postcss-custom-properties`](https://github.com/postcss/postcss-custom-properties#importfrom) and using those directly rather than [the `features` option of 'postcss-preset-env'](https://github.com/csstools/postcss-preset-env#features) which was being used previously.

@mor10 I'm not sure why but the variables in `config/cssVariables.json` were missing a double dash (e.g. `global-font-color` instead of `--global-font-color`). I had to update that as well.

mor10:
If I read the documentation directly, it should be possible to use actual `.css` files in place of the `.json` setup for the variables. That would be very nice. But don't do that just yet. I'll test the current solution on my end and if it's good, let's do the merge of this PR and then we can refine the CSS in a separate ticket. And thank you @jonathantneal for your input. Much appreciated!

mor10:
`I'm not sure why but the variables in config/cssVariables.json were missing a double dash (e.g. global-font-color instead of --global-font-color). I had to update that as well.`

@ataylorme That's how it used to work. The code originates from when I was using the now deprecated cssnext.

mor10:
@ataylorme I've tested the PR extensively, and I say **merge merge merge**.

Flag release as `v2.0-alpha` or similar and let's get all these other things on track for `v2.0` release. Excellent work!

ataylorme:
> I've tested the PR extensively, and I say merge merge merge.

💯I'll let you handle the merge and getting this into the proper branch. I am looking forward to using WP Rig v2 in my next project 😃

felixarntz:
Thanks again for all the outstanding work on this @ataylorme!!

PR #139, opened 2018-10-10
Print skip-link-focus-fix inline instead of enqueueing as deferred script

westonruter:
## Description

Copied from https://github.com/Automattic/_s/pull/1323

> Enqueueing scripts is very costly since the browser will block rendering until the script is loaded. What's more is that it is only applicable in IE11, so it is particularly bad to have this blocking script loaded for 90% of browsers which will never use it. (If IE11 supported conditional comments, those could have been used here, but it does not.) Blocking scripts should be minimized as much as possible, either by adding the `async` attribute or by outputting the `script` inline.
>
> While core doesn't yet have built-in support for `async`, core does currently add an inline script for adding the emoji logic (`print_emoji_detection_script()`/`_print_emoji_detection_script()`). A similar approach is proposed here for the `skip-link-focus-fix` script. Just like for the emoji detection script, the skip-link-focus-fix is small in size and it is only applicable to IE11.

While WPRig is using `defer` this still causes a performance penalty since it blocks `DOMContentLoaded` from firing. So printing the script inline is a better approach.

## List of changes

## Checklist:
- [ ] My code is tested.


mor10:
Also, we need to add this to the `v2.0` branch once @ataylorme is done with the refactor currently underway in #133.

westonruter:
I've redone my commit to apply to the `v2.0` branch.

dev/inc/assets.php
```diff
@@ -50,17 +50,31 @@ function wp_rig_scripts() {
...
+ * Fix skip link focus in IE11.
+ *
+ * This does not enqueue the script because it is tiny and because it is only for IE11,
+ * thus it does not warrant having an entire dedicated blocking script being loaded.
+ *
+ * @link https://git.io/vWdr2
+ */
+function wp_rig_fix_skip_link_focus() {
+	if ( wp_rig_is_amp() ) {
+		return; // See <https://github.com/ampproject/amphtml/issues/18671>.
```
hellofromtonya: 💯

bamadesigner:
Has anyone tested if we still need this IE11 fix?

mor10:
Good point Rachel

samikeijonen:
@bamadesigner Last time I checked IE11 still had this issue. But it was about a year ago.

mor10:
I think we need to keep this in because it's an accessibility issue. Even if updated IE11 is not causing a problem, it's good to keep this in place for those users who are stuck with the issue. Pushing it all the way to the end of the load chain means it'll have negligible impact on users.

westonruter:
> I think we need to keep this in because it's an accessibility issue. Even if updated IE11 is not causing a problem, it's good to keep this in place for those users who are stuck with the issue. Pushing it all the way to the end of the load chain means it'll have negligible impact on users.

This PR is not removing the fix. It's just changing how it is loaded, to prevent it from slowing down the time to `DOMContentLoaded`.

mor10:
I know. Just answering Rachel's question above.

dev/inc/assets.php
```diff
@@ -50,17 +50,35 @@ function wp_rig_scripts() {
...
+	}
+	if ( ! file_exists( get_template_directory() . '/js/skip-link-focus-fix.js' ) ) {
```
bamadesigner: Can we store `get_template_directory() . '/js/skip-link-focus-fix.js'` in a variable since we're using it twice?

mor10:
Copy that. Will merge when the refactor is done.

dev/inc/assets.php
```diff
@@ -50,17 +50,36 @@ function wp_rig_scripts() {
...
+	if ( wp_rig_is_amp() ) {
+		return; // See <https://github.com/ampproject/amphtml/issues/18671>.
+	}
+	$skip_link_focus_fix_js_file = get_template_directory() . '/js/skip-link-focus-fix.js';
+	if ( ! file_exists( $skip_link_focus_fix_js_file ) ) {
+		_doing_it_wrong( __FUNCTION__, esc_html__( 'Unable to locate js/skip-link-focus-fix.js, perhaps due to not performing a build. If not desired, please unhook wp_rig_fix_skip_link_focus from the wp_print_footer_scripts.', '2.0' ) );
+		return;
+	}
+	echo '<script>';
+	echo file_get_contents( $skip_link_focus_fix_js_file );
```
westonruter: What do you think about inlining the minified version as done in https://github.com/WordPress/twentynineteen/pull/47?

felixarntz: Let's wait until the issue there is resolved (unless it gets stuck forever), and then we can review the decided approach.

mor10:
`v2.0` was just updated (and basically rewritten). Please sync with the upstream branch, test the new code to see if your proposal is still valid, and if so update your PR.

mor10:
@westonruter We're waiting for you on this one. I assume whatever ends up happening in https://github.com/WordPress/twentynineteen/pull/47 will become de-facto best-practice, so either a) update this PR to how you think WP should do it and we'll merge it, or b) wait for a solution in Twenty Nineteen and then we'll do whatever is done there.

westonruter:
PR got merged: https://github.com/WordPress/twentynineteen/pull/47

felixarntz:
@westonruter Can you update this PR against the latest state of `v2.0`? Alternatively, you could also give me access to your fork and have me complete that work, if you prefer that (since quite a few changes have been made upstream since you opened this one up).

I suggest for our own implementation, we can as well get rid of the JS file itself completely, since it won't be used anywhere anymore. Furthermore we should only output the script if `! is_amp()`.

westonruter:
@felixarntz You should be able to push to my branch already:

image

PR #151, opened 2018-10-22
Fix comment from singular to plural and add period at end.

ecotechie:
Very basic change to copy on a comment.


felixarntz:
Looking good (test failure is caused by `master` branch itself), thanks!

PR #156, opened 2018-11-01
Update composer to match the new ruleset

dingo-d:
## Description

This PR adds the newly released [`WPThemeReview`](https://github.com/WPTRT/WPThemeReview) standard.

## List of changes

* Removed `PHPCompatibility.PHP.RemovedExtensions` because it was causing errors
* Added `WPThemeReview` ruleset
* Removed unnecessary core checks 
* Validated composer.json
* Updated composer.json with new scripts and requirements

## Checklist:
- [ ] My code is tested.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
   "name": "wprig/wprig",
   "description": "A progressive theme development rig for WordPress.",
   "type": "wordpress-theme",
-  "license": "GPL-2.0-only",
+  "license": "GPL-2.0-or-later",
   "require-dev": {
     "wptrt/wpthemereview": "^0.1.0",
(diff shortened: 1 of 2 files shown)


dingo-d:
We can specify the minimum PHP version in the `phpcs.xml.dist` using

```xml
<!-- Check code for cross-version PHP compatibility. -->
<config name="testVersion" value="7.0-"/>
<rule ref="PHPCompatibility"/>
```

So this can be remedied (or just silence errors produced by lower PHP version sniffs),

And I've modified composer license to pass composer validation, and the command was changed for simplicity, but that can be reverted 🙂

felixarntz:
Thanks for the quick response. Could you update the PR accordingly? Good catch on the license, however please use `GPL-2.0-or-later` instead. :)

Issue #86, opened 2018-08-17
Don't style classes used for structured data


benoitchantre:
Classes used for structured data should not be part of the stylesheet. Having structured data or not should not have an influence on the appearance of the site.

Related issues from Underscores:

- [Don't style hatom classes](https://github.com/Automattic/_s/issues/930)
- [Remove Microformat Classes from CSS](https://github.com/Automattic/_s/issues/1268)

mor10:
Good issue. Are you interested in contributing with a PR, or highlight which rules you'd like to see changed?

benoitchantre:
Yes. I’ll either add some comments or work on a PR next week.

benoitchantre:
I'll try to work on it this week.

benoitchantre:
At the moment, WP Rig has styles on the following classes related to [microformats](http://microformats.org/wiki/hentry):

- `hentry`
- `entry-title`
- `entry-content`
- `entry-summary`
- `updated`
- `published`

The only class not added by WP Rig is `hentry` (added by `post_class()`).

I would like to see styles removed from `hentry`, `entry-title`, `entry-content` and `entry-summary`.
This involves to add new classnames as replacements where `entry-title`, `entry-content` and `entry-summary` are used in the templates. These new classnames may depend on #109.

miklb:
I wouldn’t remove the microformat classes, just add new styling hooks

benoitchantre:
At the moment, the idea is not to remove the microformat classes, only to move styles to other classes. In a second step, structured data could be added with JSON-LD ([recommended by Google](https://developers.google.com/search/docs/guides/sd-policies)) instead of microformats.

PR #173, opened 2018-11-16
Remove style on classes used for strutured data

benoitchantre:
Having structured data or not should not have an influence on the appearance of the site.
This PR removes the dependancy of the microformats classes without changing the appearance by using new classes.

## Description

Addresses issue #86.

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/css/src/content.css
+++ b/assets/css/src/content.css
@@ -21,18 +21,18 @@
 }
 */
 
-.article {
+.entry {
 	margin-bottom: 1rem;
 }
 
 .entry-meta,
-.article-footer {
+.entry-footer {
 	font-family: var(--highlight-font-family);
 	font-size: 80%;
 }
 
 /* Hides the update date and time. */
-.date.updated:not(.published) {
+.updated:not(.published) {
 	display: none;
 }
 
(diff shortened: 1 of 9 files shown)


mor10:
I like this idea, but I think we need to move it to `v2.1` because what you've started on is a rethink of how CSS is used in the theme. This falls in line with the larger scope of moving to some form of BEM/ITCSS/CSS modules approach which we will focus on in `v2.1`. At that point we can take what you've done here and expand on it to apply to every element and style rule in the theme.

assets/css/src/content.css
```diff
@@ -22,18 +22,18 @@
 }
 */
 
-.hentry {
+article {
```
mor10: Change this to target `.article`, not the `article` element.

mor10:
We've decided to merge this, provided you fix the `article` rule at the top to `.article` and update the PR to resolve conflicts.

benoitchantre:
Thank you. Ready to merge.

mor10:
Need one more review and then we can merge.

benoitchantre:
@felixarntz The `hentry` class is added by Core to [add microformat support](http://microformats.org/wiki/hentry). If we keep `.hentry` but rename `entry-title`, which is marked as required, we will have validation errors in Google Search Console. If `hentry` is removed, then we remove support of microformat.

Do we want to remove the support of microformats?
[Google recommends to usage of schema](https://developers.google.com/search/docs/guides/intro-structured-data) (as `JSON-LD`).

felixarntz:
Thanks for clarifying. In that case I suggest we do that and modify article to entry, thus causing less unnecessary changes.

PR #164, opened 2018-11-10
Force refresh assets

benoitchantre:
## Description

Based on the comments in #39.

When WP_DEBUG is `true`, the version number is the file modification time. When WP_DEBUG is `false`, the version number is the theme version.

## List of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [ ] This pull request relates to a ticket.
- [x] My code is tested.
- [x] I want my code added to WP Rig.

PR #176, opened 2018-11-16
Get the theme version from theme object

benoitchantre:
## Description

Iteration on PR #164.

Get the theme version dynamically instead of having to update it manually. Related to comments in #39.

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/inc/assets.php
+++ b/inc/assets.php
@@ -14,7 +14,17 @@
  * @return string       Version number.
  */
 function wp_rig_get_asset_version( $file ) {
-	return WP_DEBUG ? filemtime( $file ) : '2.0.0';
+	if ( WP_DEBUG ) {
+		return filemtime( $file );
+	}
+
+	static $theme_version = null;
+
+	if ( null === $theme_version ) {
+		$theme_version = wp_get_theme( get_template() )->get( 'Version' );
+	}
+
+	return $theme_version;
 }
 
  /**


felixarntz:
I like this approach because it allows us to not define the theme version anywhere else, good call!

One thing that we should consider though is how often `wp_rig_get_asset_version()` is called. We should internally cache the call to `wp_get_theme( get_template() )->get( 'Version' )` in a static variable or something, just to ensure it's only called once. It will provide the same value on every call anyway.

benoitchantre:
Thank you for the suggestion @felixarntz. `wp_rig_get_asset_version()` uses now a static variable to remove repeated calls to `wp_get_theme()`.

felixarntz:
Looks good @benoitchantre! Minor improvement: You could even store only the version in the static variable directly. Would make the operation even faster and require less storage for the variable. :)

felixarntz:
@benoitchantre Sorry, I should have been more precise. I mean that it would be better to change your line 24 to `$theme = wp_get_theme( get_template() )->get( 'Version' );`, so that that variable only hosts the version, not the entire `WP_Theme` object. Allows to get rid of one more function call on every request.

benoitchantre:
Thank you for you code review and your advices @felixarntz.

felixarntz:
@benoitchantre Almost there - now the if clause isn't accurate anymore, it should rather check if `null === $theme_version` :)

benoitchantre:
Sorry for having missing that @felixarntz. Fixed.

felixarntz:
Awesome, thanks! No need to apologize at all!

benoitchantre:
@felixarntz One question about the PR workflow: is it ok to force push commits like I did or would you prefer to have a list of commits that need to be squased at the end?

felixarntz:
@benoitchantre Generally I recommend keeping commit history in PRs. However, for a smaller change like here it's not as crucial.

Issue #82, opened 2018-08-10
Lazy Loading Makes Custom Logo Invisible Until Scroll


joseph-farruggio:
## Issue Overview

Note: I'm only experiencing this issue on safari / iOS

Lazy Loading is preventing the custom logo from being displayed in the site header. In its place, you'll see extra vertical white space. Once you scroll, the logo appears and the header adjusts to the appropriate size.

## Steps to Reproduce

1. Ensure Lazy Loading is enabled
2. Replace your site title with a custom logo
3. View your site on iOS

mor10:
Needs further investigation. Likely culprit is how the lazy-loading script is told to ignore the site logo element.

hazephase:
I have the same issue

eddiepantoja:
This issue exists in IE as well. Thanks for all your hard work.

tessak22:
Also had this same issue, and my "Menu" link to open the mobile menu also wouldn't work. I disabled lazy loading and that seemed to resolve both issues. The menu thing was absolutely bananas because at some times, the menu would work, even when lazy loading was on, but most of the time it wouldn't work. I do not need lazy loading for this site as it's not overly image-heavy but something to note. :)

mor10:
This was on the V2 branch? Warrants further investigation.

mor10:
Closed via #198

PR #198, opened 2018-11-23
Exclude custom logo from lazy loading

mor10:
## Description
Addresses issue #82.

Excludes custom logo from lazy loading

## List of changes

## Checklist:
- [x] My code is tested.


mor10:
@felixarntz based on experimentation this seems like the simplest method. Thoughts?

mor10:
If @felixarntz puts a stamp of approval on this, I'm also going to back-port it to `v1.0.4` since it's a bug.

pluggable/lazyload/lazyload.php
```diff
@@ -214,6 +214,10 @@ function process_image_attributes( array $attributes ) : array {
 	if ( ! empty( $attributes['class'] ) && should_skip_image_with_blacklisted_class( $attributes['class'] ) ) {
 		return $attributes;
 	}
+	// Exclude custom logo from lazy loading.
+	if ( 'custom-logo' === $attributes['class'] ) {
```
felixarntz: I think we should change this to `preg_match( '/\bcustom-logo\b', $attributes['class'] )` to account for multiple classes, as right now it only matches exactly that one class being used.

Please test this to make sure it works, just in case. Then it can be merged IMO.

mor10: Tested, and works. I'm updating the PR and merging. Thanks @felixarntz!

spencerfinnell:
Is there a reason the class isn't added to the blacklist?

https://github.com/wprig/wprig/blob/master/dev/pluggable/lazyload/lazyload.php#L122

If adding a single class to that list doesn't work then there are more issues to consider!

mor10:
@spencerfinnell I believe it's called "overthinking it" from my part. This started with me trying to filter the `the_custom_logo()` or `get_custom_logo()` functions and I got lost in the weeds. Your solution makes way more sense. Thanks for pointing it out! PR updated.

mor10:
Thoughts on this update @felixarntz?

mor10:
Will update once various @felixarntz things have been merged.

felixarntz:
@mor10 This now needs to go into https://github.com/wprig/wprig/blob/v2.0/inc/Lazyload/Component.php#L281

felixarntz:
This needs an update to work, and then merge, or it needs to be punted.

mor10:
Updated to current code. Needs approval.

Issue #192, opened 2018-11-22
Allow replacing namespace in gulp task


felixarntz:
As a follow-up to #185, we should allow developers to easily replace the `WP_Rig\WP_Rig` namespace throughout.

Replacements like this one are already part of the gulp production build task, however it will currently simply replace `WP_Rig`, causing the namespace to always have the same vendor as package part in its name. However, we need to account for cases where a developer wants the namespace to be something like `The_Author\The_Theme`, or anything else (possibly consisting of more than 2 parts). Therefore I think we should make the entire `WP_Rig\WP_Rig` string replaceable with a custom namespace, in addition to the existing strings. One thing we would need to make sure though is that that specific replacement would need to fire before the other ones - specifically all instances of `WP_Rig\WP_Rig` need to already have been replaced before the `WP_Rig` instances are replaced.

This is not mission-critical for 2.0, but the sooner we have it the better.

cc @ataylorme

ataylorme:
Merged in #204

PR #204, opened 2018-11-24
Update replacement function

ataylorme:
## Description
Addresses issue #192 by adding a `PHPNamespace` option to `config/themeConfig.js` in order to allow PHP namespaces to be replaced.

PHP namespaces will be replaced first since they are the first key in `nameFieldDefaults` and `getStringReplacementTasks` runs `Object.keys` on `nameFieldDefaultsnameFieldDefaults` to generate the replacement order.

I remembered an [old comment from Tonya](https://github.com/wprig/wprig/pull/47#discussion_r201158200) about the `searchValue` setting of `string` with [`gulp-string-replace`](https://www.npmjs.com/package/gulp-string-replace) not replacing everything.

I dug into the code for the plugin and `string` [runs a normal `replace` instance](https://github.com/tomaszczechowski/gulp-string-replace/blob/1.1.2/index.js#L83), which replaces the first occurrence, not all occurrences, of the string.

This is problematic as the entire file contents are a single string and we need to replace all occurrences, not just the first.

The `regex` search method, on the other hand, [always does a global replacement](https://github.com/tomaszczechowski/gulp-string-replace/blob/1.1.2/index.js#L73).

Since the namespace string has a backslash it took some messing around with escaping as both JavaScript and regex patterns need backslashes to be escaped.

Long story short it is all sorted now and users can update the `PHPNamespace` in the config to have the namespace updated in the production build.

## List of changes
* Replace PHP Namespace with the `PHPNamespace` option in `config/themeConfig.js`
* Escape backslashes
* Move the string replace options to the getStringReplacementTasks util function

## Checklist:
- [x] My code is tested.

### DIFF
--- a/config/themeConfig.js
+++ b/config/themeConfig.js
@@ -4,7 +4,8 @@ module.exports = {
 	theme: {
 		slug: 'wp-rig', // The slug must only consist of lowercase letters, numbers and dashes.
 		name: 'WP Rig',
-		author: 'The WP Rig Contributors'
+		author: 'The WP Rig Contributors',
+		PHPNamespace: 'WP_Rig\\WP_Rig'
 	},
 	dev: {
 		browserSync: {
--- a/config/themeConfig.js
+++ b/config/themeConfig.js
@@ -5,7 +5,7 @@ module.exports = {
 		slug: 'wp-rig', // The slug must only consist of lowercase letters, numbers and dashes.
 		name: 'WP Rig',
 		author: 'The WP Rig Contributors',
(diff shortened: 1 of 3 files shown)


ataylorme:
> I approve this as it is now, but we should be aware of it. Not sure we need to add the above now, but we need to keep it in mind (in JS it would be WP_Rig\\\\WP_Rig 🤣)

Backslashes have to be escaped again for regex. It could get pretty ridiculous. I'm going to merge this for now and we can consider it if double backslashes are used in WP Rig in the future.

Issue #182, opened 2018-11-20
Add PWA plugin support


felixarntz:
We should add the `offline.php` and `500.php` templates to properly support the [PWA feature plugin](https://github.com/xwp/pwa-wp). Plus possibly minor additional tweaks if necessary.

@westonruter If you could confirm that these two templates are currently the only thing a theme developer needs to do to support that plugin, that would be great. If you have other ideas for how WP Rig might leverage the PWA plugin to succeed, please add your thoughts here.

felixarntz:
Fixed via #212.

PR #212, opened 2018-11-27
Add 500.php and offline.php templates.

felixarntz:
## Description
Addresses issue #182 

This PR adds `500.php` and `offline.php` templates to the theme, for out-of-the-box support of the PWA plugin. Having these templates provides a good foundation to developers, and they can then tweak and enhance them to their needs. Now all you need to do is activate the PWA plugin, and you're good to go in that regard.

Implementation is pretty much copying https://github.com/westonruter/twentyseventeen-westonson

Note that, along these lines, I also wrapped the navigation menu markup in a `has_nav_menu( 'primary' )` check, which was missing before. It's a good practice having such a check there in either case, here specifically though it is necessary to allow skipping the nav menu output in the `offline.php` template.

In the future, as the PWA plugin matures, we could also enhance the offline template, for example to include links to resources which have been fully cached by the service worker and are still available offline - for now I'd say the version in this PR is a good start.

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/500.php
+++ b/500.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * The template for displaying 500 pages (internal server errors)
+ *
+ * @link https://github.com/xwp/pwa-wp#offline--500-error-handling
+ *
+ * @package wp_rig
+ */
+
+namespace WP_Rig\WP_Rig;
+
+get_header(); ?>
+
+	<main id="primary" class="site-main">
+
+		<section class="error-500 not-found">
+			<header class="page-header">
+				<h1 class="page-title"><?php esc_html_e( 'Oops! Something went wrong.', 'wp-rig' ); ?></h1>
+			</header><!-- .page-header -->
+
+			<div class="page-content">
+				<p><?php esc_html_e( 'Something prevented the page from being rendered. Please try again.', 'wp-rig' ); ?></p>
+
+				<?php
+				if ( function_exists( 'wp_service_worker_error_details_template' ) ) {
+					wp_service_worker_error_details_template();
+				}
+				?>
+			</div><!-- .page-content -->
+		</section><!-- .error-500 -->
+
+	</main><!-- #primary -->
+
+<?php
+get_footer();
--- a/500.php
+++ b/500.php
@@ -19,9 +19,10 @@ get_header(); ?>
 			</header><!-- .page-header -->
 
 			<div class="page-content">
-				<p><?php esc_html_e( 'Something prevented the page from being rendered. Please try again.', 'wp-rig' ); ?></p>
-
 				<?php
+				if ( function_exists( 'wp_service_worker_error_message_placeholder' ) ) {
+					wp_service_worker_error_message_placeholder();
+				}
 				if ( function_exists( 'wp_service_worker_error_details_template' ) ) {
 					wp_service_worker_error_details_template();
 				}
--- a/offline.php
+++ b/offline.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * The template for displaying offline pages
+ *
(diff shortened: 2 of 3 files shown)


felixarntz:
@westonruter If you could briefly double-check on this, that would be great.

offline.php
```diff
@@ -0,0 +1,38 @@
...
+
+	<main id="primary" class="site-main">
+
+		<section class="error-offline not-found">
+			<header class="page-header">
+				<h1 class="page-title"><?php esc_html_e( 'Oops! It looks like you&#8217;re offline.', 'wp-rig' ); ?></h1>
+			</header><!-- .page-header -->
+
+			<div class="page-content">
+				<p><?php esc_html_e( 'The page could not be loaded because you&#8217;re offline. Please try again once reconnected.', 'wp-rig' ); ?></p>
```
westonruter: I believe this will be redundant. This message will be added by the plugin for the placeholder.

See https://github.com/xwp/twentyseventeen-westonson/blob/03934b7328f9b18dd979ae97458ad38feb6f0e1a/offline.php#L27-L33

500.php
```diff
@@ -0,0 +1,35 @@
...
+
+	<main id="primary" class="site-main">
+
+		<section class="error-500 not-found">
+			<header class="page-header">
+				<h1 class="page-title"><?php esc_html_e( 'Oops! Something went wrong.', 'wp-rig' ); ?></h1>
+			</header><!-- .page-header -->
+
+			<div class="page-content">
+				<p><?php esc_html_e( 'Something prevented the page from being rendered. Please try again.', 'wp-rig' ); ?></p>
```
westonruter: This should use `wp_service_worker_error_message_placeholder()` here instead. See: https://github.com/xwp/twentyseventeen-westonson/blob/03934b7328f9b18dd979ae97458ad38feb6f0e1a/500.php#L25-L32

felixarntz: @westonruter Ah thanks, so `wp_service_worker_error_message_placeholder()` prints a general message, while `wp_service_worker_error_details_template()` is only for showing the actual cause of the 500?

Which markup does `wp_service_worker_error_message_placeholder()` create? Just a paragraph tag with a sentence like this, or something else? Is this customizable?

Also, what happens if the site is on production and should not display the actual fatal error message for security reasons (via `ini_set( 'display_errors', 0 )`? Will `wp_service_worker_error_details_template()` automatically detect that and not output anything then?

westonruter: > Ah thanks, so `wp_service_worker_error_message_placeholder()` prints a general message, while `wp_service_worker_error_details_template()` is only for showing the actual cause of the 500?

Correct. See https://github.com/xwp/pwa-wp/blob/be8c1d702b4cf99a577b67012d09533702d4f70e/wp-includes/js/service-worker-navigation-routing.js#L67-L92

> Which markup does `wp_service_worker_error_message_placeholder()` create? Just a paragraph tag with a sentence like this, or something else? Is this customizable?

It creates a paragraph tag with a comment inside. See https://github.com/xwp/pwa-wp/blob/733563f46996ea2d5aaf52a576971f92c3748164/wp-includes/service-workers.php#L433-L438

It's not intended to be customizable.

> Also, what happens if the site is on production and should not display the actual fatal error message for security reasons (via `ini_set( 'display_errors', 0 )`? Will `wp_service_worker_error_details_template()` automatically detect that and not output anything then?

In that case, nothing will be displayed. Correct: https://github.com/xwp/pwa-wp/blob/be8c1d702b4cf99a577b67012d09533702d4f70e/wp-includes/js/service-worker-navigation-routing.js#L71-L73

lkraav:
Nice work guys. Let me know in private if I can pay somebody to do the same integration for another theme framework?

Issue #214, opened 2018-11-30
Wrong alignment for the read more link


benoitchantre:
## Issue Overview

When you add a _Read more_ block, the alignment of the `more-link` (_Continue reading_) is not aligned with the content. With Gutenberg, the `more-link` is no not enclosed in a paragraph.

Edit: Related https://github.com/WordPress/gutenberg/issues/9490

## Expected Behavior

The _Continue reading_ link should be aligned with the beginning of the content.

## Possible Solution

Set `display: block;` for the `more-link`.

## Screenshots / Video
![continue-reading](https://user-images.githubusercontent.com/1521015/49276634-56dabd80-f47f-11e8-8365-1922675791a2.png)

PR #215, opened 2018-11-30
Align .more-link with the content

benoitchantre:
## Description

Addresses issue #214 by setting this element the display property to `block` instead of `inline`.

## List of changes

## Checklist:
- [x] My code is tested.

### DIFF
--- a/assets/css/src/content.css
+++ b/assets/css/src/content.css
@@ -50,6 +50,7 @@
 .entry-content > *,
 .entry-summary > *,
 .not-found > * {
+	display: block;
 	margin-right: auto;
 	margin-left: auto;
 	padding-right: 1.5rem;
@@ -89,11 +90,6 @@
 		padding-left: 0;
 		padding-right: 0;
 	}
-}
-
-/* Align .more-link with the content. */
-.entry-content > .more-link {
-	display: block;
 }
 
 .page-links {
--- a/assets/css/src/content.css
+++ b/assets/css/src/content.css
@@ -91,6 +91,11 @@
 	}
 }
 
+/* Align .more-link with the content. */
+.entry-content > .more-link {
+	display: block;
+}
+
 .page-links {
 	clear: both;
 	margin: 0 0 1.5em;


felixarntz:
@benoitchantre Thanks for this! I prefer your second approach, accomplishing this via correct markup instead of CSS changes. Would you mind updating the PR?

benoitchantre:
@felixarntz Where should I place that code? In a new `Blocks` component?

mor10:
The link breaks out of the layout because of an omission on my part. I simply didn't test properly for this scenario. The issue is really how we want to use this function in different templates. I personally think the "Continue reading" link should be on its own line even if it's in an excerpt so I'm leaning toward the suggestion of using the `the_content_more_link` filter to wrap it in a paragraph where appropriate.

@benoitchantre I'd like your thoughts on this: Where do you think the link should be in templates where it is displayed? Inline or on a separate line?

felixarntz:
Can't we fix this in a more straightforward way, by adding `display: block;` to https://github.com/wprig/wprig/blob/v2.0/assets/css/src/content.css#L52 and https://github.com/wprig/wprig/blob/v2.0/assets/css/src/content.css#L68? We wanna make sure all child elements are block elements so that they can respect the margin/padding/max-width.

benoitchantre:
> Can't we fix this in a more straightforward way, by adding `display: block;` [...]

@felixarntz That's actually what does the commit in this PR. The CSS selector only applies when the link is not wrapped by a paragraph.

felixarntz:
@benoitchantre Yes, but the fix in this PR is very specific. I think it makes sense to fix it for _all_ content, everything that could possibly be part of that container. It's expected for all these things to be `display:block`, otherwise the rules I linked to in my comment won't have any effect, and that is why the more link for example is broken like that.

In other words: Your fix works, but I think it makes sense to fix it so that we don't run into it again with other things in the future.

benoitchantre:
> I think it makes sense to fix it for all content, everything that could possibly be part of that container.

@felixarntz Thank you for the clarification. I think there's no downside to make the fix more general as proposed.

mor10:
You have my blessing.

PR #231, opened 2018-12-09
Add optional JSON config override 

### LINKED ISSUE #228: [Enhancement] Add an optional JSON file for configuration.
I would like to check if a JSON file exists and if so, use that for the config instead of themeConfig.json.

This would solve two problems:

1) Merge conflicts when pulling upstream.
2) I'm building a 1 line theme creator and its easier to work with JSON then rewriting a JS file with find replace or parsing the AST.

Shelob9:
Fixes #228

After this commit, the themeConfig.js file, first checks if themeConfigCustom.json exists in that directory. If not, the existing config options are exported. If it does exist, the custom JSON file is used instead.

### DIFF
--- a/config/themeConfig.js
+++ b/config/themeConfig.js
@@ -1,23 +1,17 @@
 'use strict';
+const fs = require('fs');
 
-module.exports = {
+const defaultConfig = {
 	theme: {
-		slug: 'wp-rig', // The slug must only consist of lowercase letters, numbers and dashes.
+		slug: 'wprig',
 		name: 'WP Rig',
-		author: 'The WP Rig Contributors',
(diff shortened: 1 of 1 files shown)


Shelob9:
@ataylorme 
>I'm not clear on why some options are being removed in this PR.
Beacuse I used the master branch. My mistake.

>Overall I am in favor of having a default and an override configuration. However, I think we should allow a subset of options to be overridden without redeclaring options that have not changed from the default.

We could merge the JSON and the defaults right?

>I am opposed to mixing JS and JSON as the formats. I am not opposed to changing the config file format from JS to JSON.
I would rather everything was in JSON, I just didn't want to make too much of a change.

Issue #243, opened 2018-12-14
Use more template parts more granularly


felixarntz:
## Issue Overview

As my last task for 2.0 that breaks backward-compatibility by all means, I would like to improve our usage of template parts so that they are used more granularly. Some notes on what I mean by this and why it is necessary:
* With template files containing too much markup, you often end up overriding huge parts of the original codebase even though you only want to change a single line.
* Using more smaller templates rather than less larger ones also reduces the chance of merge conflicts (when updating WP Rig) and the chance of a child theme template getting out of date when the parent theme changes.
* So far WP Rig has included some functions that generate markup (such as `posted_on()` or `posted_by()`). These are opinionated in the markup they produce and thus should be easy to override - and as they solely output markup, they should instead be part of a template so that they can easily be overridden. Every markup-producing function in `inc/template-tags.php` should no longer exist and instead become part of a template file, so that `inc/template-tags.php` can just be removed.

## Related Issues and/or PRs

This issue is somewhat related to #186, as the modularization of PHP set a foundation for less template tags in use. Now, let's use templates more granularly and get rid of the remaining template tags that produce markup.

Issue #184, opened 2018-11-21
`the_attachment_navigation()` prints output even when there are no adjacent images


felixarntz:
The output should only be generated when it actually serves a purpose. Currently it will always be added even when there are no adjacent images.

Due to the expensiveness of this operation, the WP core restrictions to easily discover whether that output is needed or not, and the lack of incentive to even handle attachment pages on most sites (due to negative SEO implications, Yoast SEO disables these entirely for example), I'm even inclined to say we should get rid of this feature entirely

mor10:
Attachment pages have significant value, and being able to navigate between them is essential. The problem here, from my perspective, is lack of support from WP core and theme developers not utilizing this feature effectively. Rather than remove the feature, we should push for better support in core so the process is not so expensive.

felixarntz:
I have since reversed my opinion. Instead, the PR for granular template parts #244 actually fixes the problem that the output is unconditionally printed. :)

Also better core support is on its way hopefully (see https://core.trac.wordpress.org/ticket/45708).

PR #244, opened 2018-12-14
Use more granular template parts

felixarntz:
## Description
Addresses issue #243 (and #184 as a nice side effect)

Included changes:
* Split template parts into becoming more granular and follow a coherent structure, including usage of templates for header, content and footer:
    * Template parts for the content area are in `template-parts/content/`.
    * Template parts for the header area are in `template-parts/header/`.
    * Template parts for the footer area are in `template-parts/footer/`.
* The name of a template should be all lowercase, with underscores between words, except for template suffixes (see the second parameter of `get_template_part()`), which must use a dash per WordPress core definition. The reason for this approach is that by using underscores we prevent the chance of possible conflicts where an actual template name might collide with a template name plus the suffix.
* The most important part of this are the template parts for individual posts (all found in `template-parts/content/`):
    * `entry.php` → The overall template for a post's `article` element, plus optional additional content when in singular view.
    * `entry_header.php` → The template for the `header` element within a post's `article` element.
    * `entry_title.php` → The template for the post title within `article > header`.
    * `entry_meta.php` → The template for post meta information (such as date and author) within `article > header`.
    * `entry_thumbnail.php` → The template for the post featured image within `article > header`.
    * `entry_content.php` → The template for the actual post content within a post's `article` element.
    * `entry_footer.php` → The template for the `footer` element within a post's `article` element.
    * `entry_taxonomies.php` → The template for the associated taxonomy terms for a post within `article > footer`.
    * `entry_actions.php` → The template for additional interaction links for a post (such as link to write a comment, edit the post) within `article > footer`.
    * Each of these template files can be overridden per post type by appending the post type as suffix (as common practice in theme development anyway) - for example, if a `entry_thumbnail-page.php` is added to the same directory, it will be used instead of `entry_thumbnail.php` for posts of the post type "page". You can see this in action in the theme itself when looking at `entry-attachment.php` (thus loaded in favor of `entry.php`), which is necessary to adjust the overall output for the post type "attachment".
* Although the files have been changed to be more granular, the markup produced hasn't changed, except for the following adjustments:
    * Templates work more generally now, displaying certain areas based on whether the current post type supports them, rather than specifically assuming the current content is a "post" or "page". Some examples:
        * The date only rendered for post types that have an archive.
        * The author is only rendered for post types that support `author`.
        * The featured image is only rendered for post types that support `thumbnail`.
        * The comments template is only rendered for post types that support `comments`.
        * All associated terms for all public taxonomies associated with the current post type are rendered, which essentially ensures we have support for custom taxonomies out-of-the-box (more flexible than hard-coding output of categories and tags for the "post" post type and not considering anything else).
    * Audio and video attachment templates now properly honor whether a featured image is supported and used for them (note that core registers `thumbnail` support for `attachment:audio` and `attachment:video` specifically).
    * The navigation for previous/next attachments is now only rendered if at least one adjacent image is available.
* There is now an `error.php` template file in `template-parts/content/`, to better enforce unified markup for all error templates (such as 404, not found, empty blog page, 500, offline). For some errors, more specific templates are available, with the type of error being used as suffix (for example `error-404.php`).
* There is now a `page_header.php` template in `template-parts/content/` which essentially contains what was previously part of the `index_header()` template tag function, only with a bit more flexibility.
* The individual parts of both the theme's `header` and `footer` elements have been moved to more distinct template parts in `template-parts/header/` / `template-parts/footer` respectively.
* All old `if : else : endif;` statements in the codebase have been replaced with proper `if { } else { }` - the previous coding style is no longer preferred (by both PHP generally and WordPress specifically).

## List of changes

## Checklist:
- [x] My code is tested.


benoitchantre:
I have noticed that a lot of inline comments have been removed. I'm curious to know why? Did you find them useless?

benoitchantre:
@felixarntz Looks good to me. Thank you.

mor10:
Makes sense to me. If we're going to break the theme down into it's individual components, we might as well go all the way..

inc/Custom_Header/Component.php
```diff
@@ -69,27 +69,11 @@ public function wp_head_callback() {
...
-		else :
-			?>
-			.site-title a,
-			.site-description {
-				color: #<?php echo esc_attr( $header_text_color ); ?>;
-			}
-		<?php endif; ?>
-		</style>
-		<?php
+		if ( ! display_header_text() ) {
```
mor10: Nice cleanup. On a side note, I hate this approach to hiding content. It goes against web standards best practices and feels really clunky. Inlining CSS via PHP to hide content is crude. If we can do the same by just toggling a class name, I'd be much happier.

Issue #251, opened 2018-12-20
Search results display full posts


benoitchantre:
## Issue Overview

Without the `search.php` template, search results display full posts. This is not usefull for the end user. When there's a widget area, the developer of the theme doesn't control whether there will be a search widget or not. I think it's a good practice to keep this template enabled by default.

## Expected Behavior

Search results should always display an excerpt.

## Current Behavior

Search results display the full post because the `search.php` template is not used (in the optional directory).

## Possible Solution

Move `search.php` out of the optional directory.

mor10:
Merged. Closing.

PR #252, opened 2018-12-20
Use excerpts instead of full posts in search results

benoitchantre:
## Description

Addresses issue #251 by enabling the search template by default to display excerpts instead of full posts in search results.

Update: the search template is not enabled by default.

## List of changes

## Checklist:
- [x] My code is tested.


ataylorme:
I think this is a preference. What if we used the `rss_use_excerpt` option to determine whether to display full results or excerpts throughout the theme?

It is an option in the reading settings

> For each article in a feed, show: 
full text **or** summary

benoitchantre:
> What if we used the rss_use_excerpt option to determine whether to display full results or excerpts throughout the theme?

That may not be intuitive for site owners to use a setting aimed for feeds. If we want a setting, it should be in the customiser.

Otherwise, we could display the full post when there’s one result and the excerpt when there are many.

ataylorme:
> That may not be intuitive for site owners to use a setting aimed for feeds. If we want a setting, it should be in the customiser.

Good point. I still prefer allowing configuration over making a decision that is hard-coded one way or the other.

benoitchantre:
I’ll update the PR. What option do we want by default (full post or excerpt)?

ataylorme:
I'm indifferent. I'll let you and @mor10 sort out what the default should be

benoitchantre:
I have added a setting for Search results in Theme Options. Twenty Ninteen, Twenty Seventeen and Twenty Fifteen use excerpts for search results.

I'm enclined to change the default value or to make more changes if needed. Thank you for your feedbacks.

ataylorme:
> Twenty Nineteen, Twenty Seventeen and Twenty Fifteen use excerpts for search results.

Going with the same default as these themes makes sense to me. @mor10 what do you think?

felixarntz:
Sorry for chiming in so late, I hadn't seen this before.

I generally don't think specific Customizer controls should be added to WP Rig. This quickly goes into the direction of becoming too opinionated. WP Rig should offer defaults that theme developers can tweak, we don't need to provide controls that a user can tweak.

For things like this, I'd like to investigating another approach which would be to have separate packages that a theme developer could pull in to have such more opinionated things.

As for this PR, I think we should decide whether the default should be to show the excerpt or the content, and then let it be.

benoitchantre:
Thank you for your feedback @felixarntz. I agree we should refrain from adding controls in the Customizer.

I have noticed that since #244 has been merged, search results show full posts. Previoulsy, they were showing excerpts. Was that change intentional?

felixarntz:
@benoitchantre That must have been an oversight on my end, good catch!

However, it seems to me when looking at the old codebase that the `content-search.php` template was only ever used if there was a `search.php` template file in the theme (it is in `optional` by default).

I think we should have a vote on which of the two functions makes more sense for search results. If we decide on `the_excerpt()`, let's add a condition `if ( is_search() ) { the_excerpt(); } else { ... }` to https://github.com/wprig/wprig/blob/v2.0/template-parts/content/entry_content.php#L14

benoitchantre:
@felixarntz Yes, the `search.php` was in the optional template. My proposal is to move the `search.php` template in the theme for two reasons:

1. themes often have a sidebar, you don't know if a search widget will be used or not
1. a basic search template doesn't hurt: it's always better to have a basic search template than nothing

Let's vote in the following comments.

benoitchantre:
Vote: enable `search.php` template by default

benoitchantre:
Vote: use an excerpt instead of the full post as search results

benoitchantre:
@felixarntz I have seen you have self-assigned this pull request. Does it mean you will update the code or review the code? I'm confused.

mor10:
For reference: We assign one of the core contributors to every PR now to ensure someone is responsible for merging.  That way it doesn't get lost. Keep working on the PR, and @felixarntz is your point person.

benoitchantre:
Thank you for the clarification.

benoitchantre:
> If we decide on the_excerpt(), let's add a condition if ( is_search() ) { the_excerpt(); } else { ... } to https://github.com/wprig/wprig/blob/v2.0/template-parts/content/entry_content.php#L14

@felixarntz Why not to add the condition in `entry.php` and to add a new template part named `entry_summary.php`?
```
if ( is_search() ) {
	get_template_part( 'template-parts/content/entry_summary', get_post_type() );
} else {
	get_template_part( 'template-parts/content/entry_content', get_post_type() );
}
```

The enclosing `div` has a different class name depending its content (full or exerpt).

felixarntz:
@benoitchantre 

> Why not to add the condition in `entry.php` and to add a new template part named `entry_summary.php`?

I like that idea. Can you update the PR so it only includes these two changes?

benoitchantre:
> Can you update the PR so it only includes these two changes?

@felixarntz Yes

benoitchantre:
Here's a screenshot of the result. The first result is a page, the second a post.
Note: I have removed the post thumbnail from the search results.

capture d ecran 2019-01-22 a 21 56 57

I think we could go a bit further with improvements. Some (opiniated) ideas:

- use shorter excerpts
- remove entry taxonomies and entry actions
- add a label with the post type

With these suggested changes, the search results would be more compact and (in my opinion) more efficient for the visitor. Here's a screenshot from [Bill Erickson](https://www.billerickson.net)'s website showing these ideas.

capture d ecran 2019-01-22 a 22 01 19

Should we go further with suggestions like this? If yes, I suggest a separated pull request.
Let me know.

benoitchantre:
@flelixarntz If we wanted to filter the excerpt length, where should we add the code:

- in a new `Excerpt` component?
- in a new `Search` component?
- in the existing `Base_Support` component?
- other

I think a new `Excerpt` component would be the best place. It could contains other excerpt related filters (custom more link for example).

template-parts/content/entry_summary.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Template part for displaying a post's summar
```
felixarntz: Missing letter :)

PR #260, opened 2018-12-24
Make WP Rig child theme friendly

ataylorme:
## Description

WP Rig throws fatal errors when it is used as a child theme

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -14,13 +14,13 @@ define( 'WP_RIG_MINIMUM_PHP_VERSION', '7.0' );
 
 // Bail if requirements are not met.
 if ( version_compare( $GLOBALS['wp_version'], WP_RIG_MINIMUM_WP_VERSION, '<' ) || version_compare( phpversion(), WP_RIG_MINIMUM_PHP_VERSION, '<' ) ) {
-	require get_template_directory() . '/inc/back-compat.php';
+	require get_stylesheet_directory() . '/inc/back-compat.php';
 	return;
 }
 
 // Setup autoloader (via Composer or custom).
-if ( file_exists( get_template_directory() . '/vendor/autoload.php' ) ) {
-	require get_template_directory() . '/vendor/autoload.php';
+if ( file_exists( get_stylesheet_directory() . '/vendor/autoload.php' ) ) {
+	require get_stylesheet_directory() . '/vendor/autoload.php';
 } else {
 	/**
 	 * Custom autoloader function for theme classes.
@@ -39,7 +39,7 @@ if ( file_exists( get_template_directory() . '/vendor/autoload.php' ) ) {
 
 		$parts = explode( '\\', substr( $class_name, strlen( $namespace . '\\' ) ) );
 
-		$path = get_template_directory() . '/inc';
+		$path = get_stylesheet_directory() . '/inc';
 		foreach ( $parts as $part ) {
 			$path .= '/' . $part;
 		}
@@ -57,7 +57,7 @@ if ( file_exists( get_template_directory() . '/vendor/autoload.php' ) ) {
 }
 
 // Load the `wp_rig()` entry point function.
-require get_template_directory() . '/inc/functions.php';
+require get_stylesheet_directory() . '/inc/functions.php';
 
 // Initialize the theme.
 call_user_func( 'WP_Rig\WP_Rig\wp_rig' );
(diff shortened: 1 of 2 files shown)


mor10:
Old man winter called to tell me to tell you to take a mandatory 24 hrs off to relax and enjoy life.

ataylorme:
Says the guy checking WP Rig emails 😀

bamadesigner:
I'm going to test if WP Rig is a parent...

bamadesigner:
So. Yea. This change throws errors if WP Rig is a parent theme because its looking for the files in the child theme. :)

I'll play with this for a bit and find the best solution.

ataylorme:
@bamadesigner I wonder if we can use [`is_child_theme()`](https://developer.wordpress.org/reference/functions/is_child_theme/) to determine if the current theme is a child theme and then use `get_template_directory` or `get_stylesheet_directory` as appropriate. Although I think it will retur `true` if any child theme is activate, which would not help with WP Rig is the parent theme.

inc/Styles/Component.php
```diff
@@ -85,6 +85,18 @@ public function action_enqueue_styles() {
 			wp_enqueue_style( 'wp-rig-fonts', $google_fonts_url, array(), null ); // phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion
 		}
 
+		// Enqueue parent theme style if this is a child theme.
+		if ( is_child_theme() ) {
```
bamadesigner: 2 thoughts:

1) Isn't common practice to @import via the child theme stylesheet?
2) But more importantly: we don't have any CSS in our style.css anymore in v2.0?

functions.php
```diff
@@ -14,13 +14,13 @@
 
 // Bail if requirements are not met.
 if ( version_compare( $GLOBALS['wp_version'], WP_RIG_MINIMUM_WP_VERSION, '<' ) || version_compare( phpversion(), WP_RIG_MINIMUM_PHP_VERSION, '<' ) ) {
-	require get_template_directory() . '/inc/back-compat.php';
+	require get_stylesheet_directory() . '/inc/back-compat.php';
```
bamadesigner: If we want to make sure the WP Rig files are always called, we might have to replace all the usage of `get_template_directory()` with `dirname( __FILE__ )`.

bamadesigner:
@ataylorme Yea. I tried `is_child_theme()` but it only returns true if a child theme is active.

We might have to use some straight up PHP functions to force WP Rig files to load no matter its hierarchy.

I pushed this commit which might take care of it: https://github.com/wprig/wprig/commit/3d227893d8a75f759c8e7331f48983f40edfc96e

ataylorme:
@bamadesigner what if we put the code for both WP Rig as a child theme and WP Rig as a parent theme in `optional` and document how to use them

felixarntz:
`optional` is used for optional templates. I don't think we should put other stuff there that people _might_ use.

Specifically, as discussed in a meeting before, I'd rather document that WP Rig is intended to be used as a parent theme, and if you want to build a child theme on it, that's on you.

joseph-farruggio:
@mor10 - As a designer/developer making client sites, can you speak to how you see me using WP Rig across multiple projects?

- I'll want to version my projects in git
- I'll want to deploy my projects with git
- I'll want to pull in changes to the WP Rig repo to existing projects

I thought a WP Rig parent/child theme would solve that. I thought I would clone WP Rig to start a project and create a child theme to provide project specific changes. Having WP Rig's modern build process for each child theme is where the value is for me.

mor10:
Closing. Feel free to reopen if you have opinions.

bamadesigner:
@joseph-farruggio Hi! 

To sum up our thoughts: WP Rig works great as a parent theme but is not intended for use as a child theme. WP Rig is built to lay a solid theme foundation, which is great for a parent theme, but not for a child theme. A child theme is meant to add on or modify the foundation.



I thought a WP Rig parent/child theme would solve that. I thought I would clone WP Rig to start a project and create a child theme to provide project specific changes. Having WP Rig's modern build process for each child theme is where the value is for me.
Using WP Rig as a child theme would 

WP Rig is child theme friendly, and how you're envisioning its usage is a great method. WP Rig works great as a parent theme.

This PR title is a little confusing and the discussion was a little disjointed. Basically, the discussion revolved more around using WP Rig itself as the child theme with a different parent theme. Which is not really how its intended to be used, hence closing the PR.

bamadesigner:
@joseph-farruggio Hi!

To sum up our thoughts: WP Rig works great as a parent theme but is not intended for use as a child theme. WP Rig is built to lay a solid theme foundation, which is great for a parent theme, but not for a child theme. A child theme is meant to only add on or modify the foundation.

It sounds like you have a great scenario/need for a parent/child theme relationship. But WP Rig would serve better as the parent/the foundation. And then you can use lean child themes to simply add modifications.

I hope this helps/makes sense. I'm around if you have any questions.

Issue #278, opened 2019-01-10
Convert cssVariables.json to cssVariables.css with corresponding update to PostCSS


mor10:
`postcss-preset-env` allows for a proper CSS variables file to be used in place of a json file. See if that's possible.

mor10:
Merged. Closing.

PR #281, opened 2019-01-10
Use CSS files for custom properties and media

mor10:
Replaces `cssVariables.json` with two new files:

- `assets/css/custom-properties.css` for custom properties
- `assets/css/custom-media.css` for custom media queries

## Description

Addresses issue #278.

With this change, developers write proper CSS custom properties and custom queries and PostCSS processes them dynamically. This means developers can choose to disable the rewrite PostCSS does and instead just enqueue the new files directly to get custom properties and media working in the browser.

Because of some sort of magic, `require-uncached` is not necessary to get dynamic updates in the `npm run dev` process, which is great because `require-uncached` was coughing up a lot of errors.

## List of changes

## Checklist:
- [x] This code is tested.


gulp/constants.js
```diff
@@ -66,6 +65,8 @@ let paths = {
 		dest: `${rootPath}/`
 	},
 	styles: {
+		cssCustomProperties: `${rootPath}/assets/css/src/custom-properties.css`,
```
ataylorme: Can you use `assetsDir` instead of `rootPath`?

ataylorme:
@mor10 this looks great! One minor nitpick is all.

bamadesigner:
@mor10 I lean towards them being in the same file. There's no need for an extra file for that. They should be together.

ataylorme:
> I lean towards them being in the same file. There's no need for an extra file for that. They should be together.

I respectfully disagree but my personal preference is to modularize things. Using WP Rig for a project I would add more variables and media queries, making one monolithic file get quite large.

ataylorme:
Also previously with JSON we could pick variables and media queries from the same file and pass them separately into the corresponding parser. I'm not sure we can do that from a single CSS file.

mor10:
@ataylorme Just tested: A single file works fine.

Issue #274, opened 2019-01-07
Search results should use paginated links instead of previous / next links


benoitchantre:
## Issue Overview

Actually, when there are enough search results to display a pagination, `Older posts`/ `Previous posts` links are displayed. These labels are not optimal, because search results may contain other post types. It's also more frequent to see paginated links for search results than previous/next links (e.g. search engines). Paginated links would be better.

PR #286, opened 2019-01-13
Use numerical pagination for search results

benoitchantre:
## Description

Addresses issue #274 by replacing Older posts/Newer posts by a numerical pagination.

capture d ecran 2019-01-13 a 23 27 45

capture d ecran 2019-01-13 a 23 22 28

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/template-parts/content/pagination.php
+++ b/template-parts/content/pagination.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * Template part for displaying a pagination
+ *
+ * @package wp_rig
+ */
+
+namespace WP_Rig\WP_Rig;
+
+if ( is_search() ) {
+	the_posts_pagination(
+		array(
+			'mid_size'           => 2,
+			'prev_text'          => _x( 'Previous', 'previous set of search results', 'wp-rig' ),
+			'next_text'          => _x( 'Next', 'next set of search results', 'wp-rig' ),
+			'screen_reader_text' => __( 'Search results navigation', 'wp-rig' ),
+		)
+	);
+} else {
+	the_posts_navigation();
+}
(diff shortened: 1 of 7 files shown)


mor10:
@felixarntz wants to explore turning the navigation / pagination into a true component to make it easier for developers to add pagination options on a template-by-template or view-by-view basis.

template-parts/content/pagination.php
```diff
@@ -0,0 +1,21 @@
...
+if ( is_search() ) {
+	the_posts_pagination(
+		array(
+			'mid_size'           => 2,
+			'prev_text'          => _x( 'Previous', 'previous set of search results', 'wp-rig' ),
+			'next_text'          => _x( 'Next', 'next set of search results', 'wp-rig' ),
+			'screen_reader_text' => __( 'Search results navigation', 'wp-rig' ),
+		)
+	);
+} else {
```
felixarntz: Let's remove the if else here, and use `the_posts_pagination()` everywhere.

benoitchantre: @felixarntz Shouldn't we add a template part for post navigation named `navigation.php` where we would have `the_post_navigation()` (used in `entry.php` for singular posts) and `the_posts_navigation()`;?

felixarntz: I don't think we need it. The point of this template is to provide a single point to change that. If we had two separate templates, you would again need to change the reference to the template everywhere. The choice of `pagination` for the name is solely because we think it is more appropriate than `navigation` (because that is more generic and can mean anything), it doesn't actually mean this file _has_ to call `the_posts_pagination()`. It only means it is responsible for rendering the pagination, in whichever way the theme author prefers.

benoitchantre: Thanks you for the explanation. The code is ready to merge.

bamadesigner:
Thanks @benoitchantre!

PR #307, opened 2019-01-24
Cleanup from update CSS merge

bamadesigner:
## Description
When a couple different PRs were merged that had crossover, some commits were left out.

This PR is simply some syntax cleanup.

Addresses issue #

## List of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [ ] This pull request relates to a ticket.
- [ ] This code is tested.
- [ ] This change has been added to CHANGELOG.md
- [ ] I want my code added to WP Rig.

Issue #318, opened 2019-02-04
Allow modification of the stylesheet loading mechanism


felixarntz:
WP Rig by default preloads stylesheets and injects their `<link>` tags directly within the HTML markup and page content. This helps performance, but may not be preferred under certain circumstances. There are already checks in place to disable preloading in an AMP context, however that doesn't cover all the cases. Other scenarios where someone might potentially want to skip this is accessing the theme via the Customizer. Also, when that mechanism is disabled, all stylesheets should be enqueued rather than only registered.

Let's introduce a simple layer in the `Styles` component that allows to disable the default way of preloading for more flexibility. As part of that, let's introduce a `wp_rig()->print_styles()` template tag, which must then be used instead of directly calling core's `wp_print_styles()`. Having the method allows us to make custom tweaks (like skipping the mechanism if it's disabled) without affecting templates in the future.

mor10:
Merged. Closing.

PR #291, opened 2019-01-17
Fix wrong alignments

benoitchantre:
## Description

Addresses issue #141 by always loading `content.css`. Fixes also widgets alignments in 404.php and other small css issues.

## List of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] This pull request relates to a ticket.
- [x] This code is tested.
- [ ] This change has been added to CHANGELOG.md
- [x] I want my code added to WP Rig.

PR #319, opened 2019-02-04
Enhance/print styles

felixarntz:
## Description
Addresses issue #318

This PR allows for control about how CSS files are loaded. It still maintains the way it currently works as default, however it makes falling back to the more traditional way of printing stylesheet `<link>` tags in the head easy, as it may be preferred under certain circumstances.

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/404.php
+++ b/404.php
@@ -11,11 +11,7 @@ namespace WP_Rig\WP_Rig;
 
 get_header();
 
-/*
- * Include the component stylesheet for the content.
- * Note: If this was already done it will be skipped.
- */
-wp_print_styles( array( 'wp-rig-content' ) );
+wp_rig()->print_styles( 'wp-rig-content' );
 
 ?>
 	<main id="primary" class="site-main">
--- a/500.php
+++ b/500.php
@@ -11,11 +11,7 @@ namespace WP_Rig\WP_Rig;
 
 get_header();
 
-/*
- * Include the component stylesheet for the content.
- * Note: If this was already done it will be skipped.
- */
-wp_print_styles( array( 'wp-rig-content' ) );
+wp_rig()->print_styles( 'wp-rig-content' );
 
 ?>
 	<main id="primary" class="site-main">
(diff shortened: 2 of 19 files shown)


mor10:
This PR overlaps with #307. I suggest we adopt this one over #307 as it changes functionality. @bamadesigner, can you give this a quick once-over and see if anything from #307 should be carried over into this PR? If yes, let @felixarntz know so he can update. If no, approve this PR and close #307. Unless I'm missing something, in which case continue the convo!

felixarntz:
@mor10 @bamadesigner Ah I hadn't seen #307 - guess it's indeed covered by this one too.

bamadesigner:
In testing this PR, I discovered an issue.

Basically, there's no co-relation to when `print_styles()` is used inside a template and what CSS files are pre-loaded. The logic that preloads a CSS file is "hard coded" inside the `get_css_files()` function.

For example: I could decide I don't want to load the content.css file and remove `'wp-rig-content'` from the `print_styles()` call in the front-page template but the content.min.css file is still pre-loaded because its logic is always set to true in the `get_css_files()` function.

Since we're already mucking around in here, it would be great to address the issue in this PR.

bamadesigner:
FYI: #291 adds a usage of `print_styles()`. @benoitchantre

felixarntz:
@bamadesigner 
> Basically, there's no co-relation to when `print_styles()` is used inside a template and what CSS files are pre-loaded. The logic that preloads a CSS file is "hard coded" inside the `get_css_files()` function.
>
> For example: I could decide I don't want to load the content.css file and remove `'wp-rig-content'` from the `print_styles()` call in the front-page template but the content.min.css file is still pre-loaded because its logic is always set to true in the `get_css_files()` function.

I agree this could use some additional thoughts. However, this shortcoming is not trivial to fix and is not introduced by this PR, so I think it deserves a separate issue and PR. Since outputting the styles is decided after they're being preloaded, it'd be rather complex to address this. Could you create a separate issue for that for "Future Release"?

PR #324, opened 2019-02-07
Don’t compile custom CSS files

bamadesigner:
## Description
The CSS compile was compiling the CSS variable property and media files and creating unnecessary, empty files. This removes them from compilation.

Addresses issue #

## List of changes

## Checklist:
- [ ] This code is tested.
- [ ] I want my code added to WP Rig.


mor10:
If `"preserveCSSVars": true`, the compilation needs to be different. Ideally the contents of `custom-properties.css` and `custom-media.css` need to be added to the top of `style.css` or enqueued in `head` so the variables are included in the document.

ataylorme:
> If `"preserveCSSVars": true`, the compilation needs to be different. Ideally the contents of `custom-properties.css` and `custom-media.css` need to be added to the top of `style.css`

@mor10 appending the contents of `custom-properties.css` and `custom-media.css` to the top of `style.css` when creating the production bundle is straightforward. What are your thoughts on getting them loaded when gulp is actively running a build? `style.css` isn't processed/watched during the build/dev phase. Maybe they can be added to `global.css`? That presents a problem though as `global.css` can be removed

mor10:
In conversation, @ataylorme and I realized because PostCSS provides automatic fallback in CSS for css custom properties, we don't actually need the `preserveCSSVars` logic. Since that's the case, we can take that logic out and inject `custom-properties.css` and `custom-media.css` into `global.css` via `@import` and the [postcss-import](https://github.com/postcss/postcss-import) plugin which parses those rules into clear CSS.

bamadesigner:
PostCSS issues addressed in #335. This now covers simply ignoring src only files from the build.

Issue #238, opened 2018-12-12
Update editor styles to new convention


mor10:
See https://wordpress.org/gutenberg/handbook/designers-developers/developers/themes/theme-support/#editor-styles

PR #332, opened 2019-02-15
Update editor styles to new convention

benoitchantre:
## Description

Addresses issue #238 by updating the editor stylesheet and how it is loaded.

## List of changes

## Checklist:
- [x] This code is tested.


benoitchantre:
@felixarntz, @mor10 Could we move the loading of the editor stylesheets in the _Editor_ component?

inc/Styles/Component.php
```diff
@@ -174,22 +174,15 @@ public function action_preload_styles() {
 	}
 
 	/**
-	 * Enqueues WordPress theme styles for the block editor.
+	 * Enqueues WordPress theme styles for the editor.
 	 */
-	public function action_enqueue_block_editor_styles() {
+	public function action_add_editor_styles() {
```
mor10: I'd like a sanity check on this from @felixarntz.

benoitchantre:
Build failed for an unrelated reason to this PR:

> PHP Fatal error:  Cannot declare class PHPUnit_Util_Test, because the name is already in use in /tmp/wordpress/tests/phpunit/includes/phpunit6/compat.php on line 18

inc/Styles/Component.php
```diff
@@ -174,22 +174,15 @@ public function action_preload_styles() {
...
 
 		// Enqueue Google Fonts.
-		$google_fonts_url = $this->get_google_fonts_url();
-		if ( ! empty( $google_fonts_url ) ) {
```
felixarntz: This check needs to be maintained. Only if Google fonts URL is not empty, it should be added as editor style.

inc/Styles/Component.php
```diff
@@ -74,7 +74,7 @@ public function get_slug() : string {
 	public function initialize() {
 		add_action( 'wp_enqueue_scripts', array( $this, 'action_enqueue_styles' ) );
 		add_action( 'wp_head', array( $this, 'action_preload_styles' ) );
-		add_action( 'enqueue_block_editor_assets', array( $this, 'action_enqueue_block_editor_styles' ) );
+		add_action( 'admin_init', array( $this, 'action_add_editor_styles' ) );
```
felixarntz: I'm curious why you chose `admin_init` here. Typically, `add_editor_style()` is called on `after_setup_theme`.

inc/Styles/Component.php
benoitchantre: Thanks

inc/Styles/Component.php
benoitchantre: I have probably looked at [this example](https://developer.wordpress.org/reference/functions/add_editor_style/#comment-1171). Changed to `after_setup_theme` following your recommendations.

benoitchantre:
@felixarntz It should be good to merge now.

mor10:
Great work @benoitchantre

benoitchantre:
Thank you @mor10.

ataylorme:
#356 needs to be merged before this one

benoitchantre:
Blocked by #361.

ataylorme:
@benoitchantre did `@import "../custom-properties.css";` solve the issue in #361?

benoitchantre:
> @benoitchantre did `@import "../custom-properties.css";` solve the issue in #361?

@ataylorme No. I have merged `v2.0` and imported `custom-properties.css` so that everybody can easily check the issue described in #361. The workaround I have found is to manually rename `:root` to `body` in `editor-styles.min.css`. This step could be done by the build process.

benoitchantre:
On hold for a decision on #361.

mor10:
I suggest we merge these related PRs in the following order:

- #367 
- #364 
- #332 (this PR)

My testing indicates the two first PRs resolve the editor styles blocker for this PR.

benoitchantre:
> My testing indicates the two first PRs resolve the editor styles blocker for this PR.

The issue with custom properties not working in the editor only happen once your enqueue the editor stylesheet with `add_editor_style()` as recommended by the Gutenberg Handbook. The styles are processed, but custom properties set to `:root` are transformed to `.editor-styles-wrapper :root` (doesn't work). The solution is to rename `:root` to `html` if we want to preserve custom properties.

mor10:
Riiiiiiiiiight... I missed that part. OK. I'll do some more testing to see if there's a way around breaking the  CSS markup.

mor10:
I tested it and you're right @benoitchantre, the other PR doesn't solve the problem. What the core editor is doing here is extremely heavy handed and stands in the way of good coding practices.

The quick and dirty solution is indeed to replace `:root` with `html` in `custom-properties.css` even though that's poor form. @ataylorme is it possible to do this transform during the build process (swapping the selector name), or do we have to dumb down `custom-properties.css` to comply with Gutenberg doing things wrong?

ataylorme:
>  @ataylorme is it possible to do this transform during the build process (swapping the selector name)

It may be _possible_ by editing the contents of the file in the middle of the gulp stream, but it is not something I want to do, think we should do, or am comfortable doing.

mor10:
That's what I thought @ataylorme. Some digging in the Gutenberg issue pile surfaced this, which was closed without resolution: https://github.com/WordPress/gutenberg/issues/11955
I've filed a new issue to highlight the need to get this resolved properly: https://github.com/WordPress/gutenberg/issues/14297

ataylorme:
What if we went back to `enqueue_block_editor_assets` instead of `add_editor_style` until the bug in Gutenberg is fixed?

mor10:
That's also doing it wrong. I'm trying to figure out a workaround.

ataylorme:
> I suggest we merge these related PRs in the following order:
> #367
> #364
> #332 (this PR)

I merged #367 and #364. See [this comment](https://github.com/wprig/wprig/pull/364#issue-258362877) for what changed in 364. I'm not sure what the workaround is. If `preserve` needs to be changed then this PR can be updated to do so in `config/config.default.json`.

benoitchantre:
@ataylor Would it be complicated to have in the build task a separated setting (or hard coded value) for the editor stylesheet, with a default value for `preserve` set to `false`?

mor10:
@benoitchantre Test this along with #369 and see if it works for you. It works on my end, but I need to make sure I'm not missing something.

For reference, the mishandling of `:root` in editor styles has been patched in Gutenberg and is slated for WordPress 5.2 release which will happen in April or May I think: https://github.com/WordPress/gutenberg/pull/13325

#369 is a temporary workaround allowing developers to do things correctly today, and then we can revert gracefully when WP core gets updated.

Issue #290, opened 2019-01-17
front-page.php should not load a sidebar


benoitchantre:
## Issue Overview

Introduced with #244.

I’ll work on it as soon as #291 is merged.

mor10:
For reference, `front-page.php` loads a sidebar to provide an example of how to load a sidebar in a non-standard way.

benoitchantre:
I just checked again `master` and see no sidebar. Comments are displayed as a second column when they are enabled.

In `v2.0`, I see the same as `master` plus a sidebar as a third column.
Is it how it should be or a small mistake introduced when refactoring all template parts?

mor10:
Sorry @benoitchantre. My bad. I was testing for the wrong scenario. You are absolutely right, there should not be a sidebar in the `front-page.php` template. 

The template was mainly designed to handle the blog index, not a static page, and I see in testing I didn't account for the static page as front page scenario. I'm going to create a PR to fix these issues and update the overall styling (which is hideous now that I look at it).

PR #338, opened 2019-02-28
Update front page

mor10:
## Description

Addresses issue #290

- Removes call to sidebar in `optional/front-page.php`
- Marginally improves the styling of the front page content when the template is used to display the blog index. 

## Todo
Get a designer to come up with a better look for the front page template.

## List of changes

## Checklist:
- [x] This code is tested.


optional/front-page.php
```diff
@@ -27,5 +27,5 @@
 		?>
 	</main><!-- #primary -->
 <?php
-get_sidebar();
+
```
bamadesigner: Not a big deal but none of the other files have this extra line.

bamadesigner: In Sidebars/component.php, its hard-coded to add "has-sidebar" selector to body class.

mor10: No, the code in `Sidebars/component.php` does the opposite: it excludes the `has-sidebar` selector from this template. Unless I'm completely misreading the logic:

https://github.com/wprig/wprig/blob/v2.0/inc/Sidebars/Component.php#L90-L92

mor10:
where are we on this?

benoitchantre:
I’ll test that before Monday.

benoitchantre:
@mor10 When comments are enabled, they are displayed next to the content, instead of below it.

optional/front-page.php
bamadesigner: You're right.

mor10:
Ah. I see now. This template was originally designed to replace the front page as blog index, not the front page when set to display a single page. I need to figure out if I can hook into that logic. The grid only makes sense when the front page displays the blog index.

mor10:
Updated PR with a conditional to test what is being displayed on the front page. With this update, the front page uses a grid only when the index is displayed. Please test and confirm so we can merge.

benoitchantre:
Works for me.

Is there a reason to keep `front-page.css`? This file is not used on front-end anymore.

bamadesigner:
Works for me. @benoitchantre, we need the front-page.css if y'all want these blog index styles that are showing up on the front-end.

bamadesigner:
I don't see the comment issue.

mor10:
@felixarntz do you want to resolve your complaint so we can merge?

mor10:
Reviewing @felixarntz's comment, I think we're good. Merging.

Issue #346, opened 2019-03-02
Blockquotes in Classic Editor are not styled


benoitchantre:
## Issue Overview

We should have consistent styles between elements that are in the Classic Editor and their block version.

**Here's how is displayed a Blockquote in the Classic Editor.**
<img width="783" alt="blockquotes - wprig" src="https://user-images.githubusercontent.com/1521015/53677995-188cae80-3cb9-11e9-8251-18c3e21aaf33.png">
 
**Instead of something like that**
<img width="738" alt="blockquote block" src="https://user-images.githubusercontent.com/1521015/53678009-45d95c80-3cb9-11e9-944c-a3f1661b3371.png">

**Same test with _Twenty Nineteen_ (consistent with their Blockquote block style)**
<img width="705" alt="blockquotes - twentynineteen" src="https://user-images.githubusercontent.com/1521015/53677997-1e828f80-3cb9-11e9-921d-b977cc0a8134.png">

PR #353, opened 2019-03-04
Style blockquotes like the quote block

benoitchantre:
## Description

Addresses issue #346 by setting the styles from the default Quote block to `<blockquote>` and `<cite>`.

Result:
capture d ecran 2019-03-04 a 14 09 47

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/assets/css/src/global.css
+++ b/assets/css/src/global.css
@@ -128,7 +128,17 @@ i {
 }
 
 blockquote {
+	border-left: 4px solid #000;
 	margin: 0 1.5em;
+	padding-left: 1em;
+}
+
+blockquote cite {
+	color: #6c7781;
+	font-size: 13px;
+	margin-top: 1em;
+	position: relative;
+	font-style: normal;
 }
 
 address {
--- a/assets/css/src/global.css
+++ b/assets/css/src/global.css
@@ -129,11 +129,18 @@ i {
 
 blockquote {
 	border-left: 4px solid var(--color-quote-border);
-	margin: 0 1.5em;
+	margin: 0;
 }
 
 .entry-content > blockquote {
 	padding-left: 1em;
+	margin-left: 1.5rem;
+}
+
+@media (--content-query) {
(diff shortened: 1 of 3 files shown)


assets/css/src/global.css
```diff
@@ -128,7 +128,17 @@ i {
 }
 
 blockquote {
+	border-left: 4px solid #000;
```
mor10: Thinking these colors should be variables to be consistent.

assets/css/src/global.css
```diff
@@ -128,7 +128,17 @@ i {
...
 	margin: 0 1.5em;
+	padding-left: 1em;
+}
+
+blockquote cite {
+	color: #6c7781;
```
mor10: Thinking these colors should be variables to be consistent.

assets/css/src/global.css
benoitchantre: In this case, we replicate the style from _Quote_ Blick defined in a external stylesheet. Using a variable will not make our styles consistent a unless we overload the border color also for the _Quote_ block. I think add a variable and overload the border left color for the  _Quote_ block.

assets/css/src/global.css
benoitchantre: In this case, we replicate the style from _Quote_ Blick defined in a external stylesheet. Using a variable will not make our styles consistent a unless we overload the border color also for the _Quote_ block. I think add a variable and overload the border left color for the  _Quote_ block.

benoitchantre:
I'll add some styles in the editor styles, either in this PR or in #332, depending on what will be merged first.

ataylorme:
@benoitchantre does this address #362 as well?

benoitchantre:
> does this address #362 as well?

@ataylorme No, this only makes the styles of a Quote inserted in Classic Editor identical to the _Quote_ block. #362 needs decision how to fix it. You can vote for one of the proposed options or propose another one.

Issue #355, opened 2019-03-04
Custom properties values without units are not evaluated


benoitchantre:
## Issue Overview

Values in for custom properties that have no units are not working properly.

Browsers dont't evaluate values like `var(--content-width) rem`.
Tested with Firefox and Chrome.

ataylorme:
Closed by #356

PR #356, opened 2019-03-04
Fix custom properties without units not evaluated

benoitchantre:
## Description

Addresses issue #355.

@ataylorme Not sure when stylelint has to be disabled.

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/assets/css/src/custom-properties.css
+++ b/assets/css/src/custom-properties.css
@@ -11,7 +11,7 @@
 	--global-font-size: 20;
 	--global-font-line-height: 1.4;
 	--highlight-font-family: "Roboto Condensed", sans-serif;
-	--content-width: 45;
+	--content-width: 45rem;
 	--dropdown-symbol-width: 0.7em;
 
 	/* Custom editor colors */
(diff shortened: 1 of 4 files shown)


benoitchantre:
To merge before #332 if possible (#332 will need to be updated).

ataylorme:
> Not sure when stylelint has to be disabled.

Neither am I. I can trace the stylelint disables back to the giant v2 merge in #133, though that just renamed files and didn't change the contents. @mor10 any idea where the stylelint disables came from, why they were added, or if they are still needed?

I think cleaning them up (or removing them) is something we can do in a different issue though and shouldn't block this from merging

assets/css/src/custom-properties.css
```diff
@@ -32,8 +32,8 @@
 	--color-link-active: #00a0d2;
 
 	/* Custom editor font sizes */
-	--font-size-small: 16;
-	--font-size-regular: 20;
-	--font-size-large: 36;
-	--font-size-larger: 48;
+	--font-size-small: calc(16 / var(--global-font-size) * 1rem);
```
mor10: Love this.

Issue #314, opened 2019-01-31
Allow skipping translation file generation


felixarntz:
Currently, every build will generate a new `.pot` file to use for writing translations. However, that might not always be desired. Particularly as themes in the wordpress.org repository don't require or use these, as wordpress.org has its own translation system that themes automatically integrate with.

I think it would be great to have a setting in our `config.json`, called `useLanguagesDir`, that defaults to true (as it is now), but can be set to false. In that case, the `translate` tasks in Gulp should not run.

The other, a bit more challenging thing this would involve is to dynamically change https://github.com/wprig/wprig/blob/v2.0/inc/Localization/Component.php#L47 to only `load_theme_textdomain( 'wp-rig' )`. A potential solution would be to simply keep the code in the PHP file as is, and add that part as another PHP replacement item when building the theme.

@ataylorme Pinging you to get your feedback.

ataylorme:
@felixarntz for the gulp side I propose we remove the translation process from the default tasks. We can expose it as a stand-alone task through npm script (`npm run pot` or similar).

The reason for this is that, as you say, people may wish to create translation files in different ways (or not at all). I don't think we should lock people into a default.

Personally, I would want to [make them with wp-cli](https://github.com/wp-cli/i18n-command) so that PHP and JavaScript files are processed and will do this in a build step on a CI server.

felixarntz:
I like the idea of excluding the task from the default ones.

mor10:
Can you two sort this out and make an executive decision. I don't have an opinion on the matter either way as long as translation files _can_ be made.

ataylorme:
> I don't have an opinion on the matter either way as long as translation files can be made.

Currently, the translation process is run on *source* files, not *production* files. @mor10 and @felixarntz what do you think of having the current translation task (`npm run translate`) run `gulp-pot` on the production directory?

If we are no longer creating the `.pot` files on each dev run I can't think of a reason to keep basing them off of the source files and not the production ones.

mor10:
that makes sense, especially considering production files may look different from development files at some point in the future.

ataylorme:
Cool. I'll add a configuration option, as @felixarntz suggested in order to disable the `.pot` file generation on production builds.

ataylorme:
Closed by #358

PR #358, opened 2019-03-04
Pot on prod build

ataylorme:
## Description

Addresses issue #314.

* Removes translation file generation from development builds
* Adds a new `export:generatePotFile` configuration variable to control `.pot` file generation on production builds
* Updates the Localization Component to only run `load_theme_textdomain` if `.pot` files exist
* Updates `README` documentation

## List of changes

## Checklist:
- [x] This code is tested.


ataylorme:
@felixarntz I would like to get your thoughts on the changes to the localization component.

inc/Localization/Component.php
```diff
@@ -30,7 +61,10 @@ public function get_slug() : string {
 	 * Adds the action and filter hooks to integrate with WordPress.
 	 */
 	public function initialize() {
-		add_action( 'after_setup_theme', array( $this, 'action_load_textdomain' ), 1 );
+		// Only add textdomain actions if there are .pot files.
+		if ( $this->pot_files_exist() ) {
```
felixarntz: @ataylorme Just double-checking, is it really okay to run this logic on every page load?

ataylorme: @felixarntz if we don't do the check then `action_load_textdomain` could run without any valid files. Is that acceptable? If not, can you provide an alternative solution?

mor10: Might be a dumb idea, but would caching help here?

ataylorme: I think storing the result of `pot_files_exist` in a cache, like a transient, would be less efficient than just running the function

bamadesigner: WordPress Core has logic that checks if the file is readable/exists and, if not, doesn't load so we don't need this logic.

bamadesigner:
Thanks @ataylorme!

Issue #363, opened 2019-03-05
watch task fails


benoitchantre:
## Issue Overview

When you run `npm run dev`, it stops at the watch task with the following error:

```
TypeError: expected a string
```

The error is located in line 49.

To fix this error, we can replace `[paths.styles.src, paths.styles.cssCustomProperties, paths.styles.cssCustomMedia]` by `paths.styles.src`, but doing that, a change in `custom-properties.css`, `custom-media.css` or `reset.css` will not trigger the style task because they are excluded.

Cc. @ataylorme

ataylorme:
@benoitchantre can you test #364 and see if solves the issue for you?

ehowey:
Is the same as what I had happen here: https://github.com/wprig/wprig/issues/340

benoitchantre:
> Is the same as what I had happen here: #340

Yes

ataylorme:
@benoitchantre can you verify this is fixed for you? If so, please close the issue.

benoitchantre:
✔️ works

PR #364, opened 2019-03-05
Don't exclude imported files in CSS source path

ataylorme:
## Description

Addresses issue #363 by removing the exclude for currently imported CSS files (`reset.css`, `custom-properties.css`, and `custom-media.css`) in the CSS source path. The excluded files have been moved to a configuration option and are added during the styles task, not watch tasks.

The excluded files are not processed and minified as production files, which was the original goal of having them excluded. However, I prefer them in config rather than being hard-coded.

Config options for `customProperties`, `customMedia` and `preserve` have also been added. They directly effect the options passed to postCSS without hard-coding file paths or requiring users to edit gulp files.

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/config/config.default.json
+++ b/config/config.default.json
@@ -13,7 +13,7 @@
       "https": false
     },
     "styles": {
-      "preserve": false,
+      "preserve": true,
       "customProperties": [
         "custom-properties.css"
       ],
--- a/config/config.default.json
+++ b/config/config.default.json
@@ -13,6 +13,7 @@
       "https": false
     },
     "styles": {
+      "postCSSPreserve": true,
       "ignoredSourceFiles": [
         "reset.css",
         "custom-properties.css",
--- a/config/config.default.json
+++ b/config/config.default.json
@@ -13,7 +13,7 @@
       "https": false
     },
     "styles": {
-      "preserve": true,
+      "preserve": false,
       "customProperties": [
         "custom-properties.css"
       ],
--- a/config/config.default.json
+++ b/config/config.default.json
@@ -12,6 +12,16 @@
       "bypassPort": "8181",
       "https": false
     },
+    "styles": {
+      "ignoredSourceFiles": [
+        "reset.css",
+        "custom-properties.css",
+        "custom-media.css"
+      ]
+    },
+    "scripts": {
+      "ignoredSourceFiles": []
+    },
     "debug": {
       "styles": false,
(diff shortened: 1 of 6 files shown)


mor10:
@bamadesigner and @felixarntz had strong opinions on this.

benoitchantre:
Works for me. Thanks @ataylorme.

ataylorme:
@mor10 I updated the PR code and the description. The files are back to being excluded but are now defined in config, rather than being hard-coded in the gulp files.

mor10:
With these new updates, combined with the `editor-styles.css` fix in #367, and with `preserve: true` reinstated, I am happy with this.

Custom properties and media should be passed as-is through to the final code unless the user explicitly turns this feature off. The fallbacks provided by PostCSS are sufficient, and with the fix in #367 work across the front- and back-end.

config/config.default.json
```diff
@@ -12,6 +12,23 @@
       "bypassPort": "8181",
       "https": false
     },
+    "styles": {
+      "preserve": false,
```
mor10: Set to `true`.

benoitchantre:
> The fallbacks provided by PostCSS are sufficient, and with the fix in #367 work across the front- and back-end.

I have merged this branch locally into #332 (local version) and set `preserve` to `true`, but it doesn't work in the editor.

ataylorme:
@benoitchantre did you merge #367 locally as well? I think that PR is needed too in order to fix the editor.

In either case, I would like to merge this to fix the broken watch task in `master`. If we need to flip the default config value for `preserve` again we can do that in a follow up pull request.

benoitchantre:
> @benoitchantre did you merge #367 locally as well?

Yes. This PR only adds `@import "../custom-media.css";` which is not (yet) required.
`custom-properties.css` was already imported with the latest changes.

Here's what I have when I open the editor.

With `preserve` set to `false`:
capture d ecran 2019-03-06 a 20 55 48

With `preserve` set to `true`:
capture d ecran 2019-03-06 a 20 55 12

benoitchantre:
To get the right preview in the editor with `preserve` set to `true`, `:root` needs to be renamed to `html`.

ataylorme:
> In either case, I would like to merge this to fix the broken watch task in `master`.

Moving ahead with merging.

Issue #361, opened 2019-03-05
Custom properties are missing in the editor 


benoitchantre:
## Issue Overview

Trying to update #332 with the latest changes on `v2.0`, I have noticed that custom properties are not available in the editor even if we import them in `editor-styles.css`. The styles declared in `editor-styles.css` are processed by the block editor and used as inline. We loose our custom properties during that process.

## Possible Solution

Add fallback for custom properties.

benoitchantre:
@ataylorme Were the custom properties fallback removed intentionnaly?
[To support IE11, we need fallbacks](https://caniuse.com/#feat=css-variables) according to caniuse.com

ataylorme:
The variables are imported into the global CSS. I didn't think of the case where editor styles are used stand-alone

> I have noticed that custom properties are not available in the editor even if we import them in editor-styles.css

Are you saying that `@import "../custom-properties.css";` at the top of the `editor-styles.css` files does not work?

If so, that is likely due to [`postcss-import` skipping duplicate imports](https://github.com/postcss/postcss-import#skipduplicates).

cc @mor10

benoitchantre:
> Are you saying that `@import "../custom-properties.css";` at the top of the `editor-styles.css` files does not work?

Custom properties can be found in `editor-styles.min.css`, but this file is then processed by the block editor and not loaded. That process renames some selectors and inlines the styles. This is when we loose our custom properties values.

I think the best solution would be to rename `:root` to `body` in `editor-styles-min.css` (I have tested that manually, it works). An alternative would be to add fallbacks for custom properties.

samikeijonen:
@benoitchantre Did you notice that Core had prefixed `:root` with `.editor-styles-wrapper :root`?

mor10:
This is a bug in the core editor and is going to become an issue down the road. Might be a good cause for a ticket to core or in the Gutenberg repo.

mor10:
@benoitchantre I just tested, and if I add the following to `editor-styles.css`, custom properties are added to the top of the minified editor stylesheet and things work as expected:

```
@import "../custom-media.css";
@import "../custom-properties.css";
```

benoitchantre:
It seems that I haven't been clear enough that this issue has to be seen in the context of #332. Correct me if I'm wrong.

@mor10 I have updated #332 in order to facilitate tests. Can you check again?

> @benoitchantre Did you notice that Core had prefixed `:root` with `.editor-styles-wrapper :root`?

[...]

PR #368, opened 2019-03-07
Automatically transform editor CSS with custom media and custom properties

ataylorme:
## Description

Addresses issue #361. This uses [`gulp-tap`](https://www.npmjs.com/package/gulp-tap) to read CSS file contents in the middle of the stream during the `styles` task. 

If the file's path, relative to the CSS source directory, starts with `editor/` then utilities from [`postcss-custom-utils`](https://www.npmjs.com/package/postcss-custom-utils) are used to apply the custom media and custom properties to the file contents and the modified source file contents continue down the stream.

[This article](https://gulpjs.org/recipes/make-stream-from-buffer.html) suggests it might be possible to be more efficient by editing the file contents in memory.

The `@import` statements in the editor CSS source files for `custom-media.css` and `custom-properties.css` were removed as they are now applied directly to the file contents during the build process.

This was mostly a learning exercise for me to see if I _could_ do this. I still don't think we _should_ do this.

I think it might be easier to dynamically change `preserve` to `false` for editor CSS source files or maybe even use a different task all together for editor CSS source files

Maybe @jonathantneal is willing to help us find a solution to change the `:root` selector to `html` for specific CSS source files in the gulp task?

## List of changes

## Checklist:
- [x] This code is tested.


mor10:
Impressive! I need to think on this one for a bit.

mor10:
Oops. Clicked the wrong button 😑

jonathantneal:
There is a plugin to swap `:root` for `html`. https://github.com/jonathantneal/postcss-unroot

ataylorme:
I'm closing this because we shouldn't take this approach.

PR #369, opened 2019-03-07
Dynamically set postcss preserve to false for editor CSS

ataylorme:
## Description

Addresses issue #361 by adding a new `editorStyles` gulp task for editor CSS with `preserve` forced to `false` and #369 by making sure the `preserve` value in config is respected, outside of the previous exception for editor files.

This should be a temporary change as [the `:root` issue in Gutenberg was fixed](https://github.com/WordPress/gutenberg/issues/14297#issuecomment-470511862) and will be included in WordPress `5.2`, which is currently scheduled for April 30th, 2019.

The following additional cleanup is also made in this PR:
* Rename `@import`ed CSS files so that they are prefixed with an underscore
* CSS and JS source files prefixed with an underscore `_` are considered partials and are not built to production.
* Added `@import "_custom-media.css";` and `@import "_custom-properties.css";` to all CSS source files.
* Removed the `ignoredSourceFiles` option from `config.default.json`
  * Partial files prefixed with `_` are automatically ignored
* Added `stage` and `feature` options to `config.default.json` so that users may override the default `postcss-preset-env` options.
* Remove the `postcss` `postcss-custom-properties` and `postcss-custom-media` plugins
  * Configure `postcss-preset-env` properly for custom properties and custom media
* Opt-in to the `custom-media-queries`, `custom-properties`, and `nesting-rules` features of `postcss-preset-env` by default.
* Add `cleanCSS` and `cleanJS` tasks which remove all built/minified files of their respective type.
  * This is run before re-building so that deleted files are removed.
* Add a `gulp` alias to `npm scripts` so that `npmr un gulp -- <task name>` can be called to invoke any single gulp task as a one-time operation.

## List of changes

## Checklist:
- [x] This code is tested.


benoitchantre:
I have found some differences on front-end. We don't have the fallbacks for custom properties anymore and we have a media query that doesn't work.

Before (`v2.0`):
capture d ecran 2019-03-07 a 20 51 45

Now (this PR):
capture d ecran 2019-03-07 a 20 52 36

When I merge this PR into #332, I get the fallbacks for custom properties, but no custom properties. This doesn't work for the browser: it can't handle undefined values.

capture d ecran 2019-03-07 a 21 03 49

@mor10 Did I miss something in my tests?

Here's what I did:
1. `npm install`
1. `npm run build`
1. inspect styles

ataylorme:
@benoitchantre those steps are correct. I am not seeing the same issue (screenshot below) when testing just this branch (I didn't merge in #332).

Could there be some conflict with the two or perhaps a bug in 332?

screen shot 2019-03-07 at 4 39 22 pm

benoitchantre:
@ataylorme It seems that we have the same result. If you switch branch to `v2.0` and run `npm run build` again, you will see that you had fallbacks for custom properties (first screenshot from my previous comment). I think the latest changes in this PR shouldn't remove these fallbacks.

My first test (front-end) was on this PR, without any merge into another branch.
My second test was with this PR merged into 332 to check the editor styles.

ataylorme:
> It seems that we have the same result. I think the latest changes in this PR shouldn't remove these fallbacks.

@mor10 specifically wanted `preserve` set to `true` by default.

My browser is processing the CSS variables and not relying on the fallback, so things look fine.

@benoitchantre does setting `preserve` to `false` in `config.json` work for you? When I do that both the fallback and custom properties are printed.

It seems like this may be a personal preference/configuration, which is the aim of `config.json`, to allow the defaults in `config.default.json` to be overwritten on a per-project basis.

P.S. you can use `npm run build -- styles` to run just the styles task for testing.

benoitchantre:
> @mor10 specifically wanted preserve set to true by default.

I'm not against that choice, but it seems the behaviour of the build process has changed and the results are not following what we should get (it was working as described below in `v2.0`)

Example from the [documentation](https://github.com/postcss/postcss-custom-properties) :
```css
:root {
  --color: red;
}

h1 {
  color: var(--color);
}
```

with `preserve`set to `true`, we should have
```css
:root {
  --color: red;
}

h1 {
  color: red;
  color: var(--color);
}
```

with `preserve`set to `false`, we should have
```css
h1 {
  color: red;
}
```

From your screenshot, I see the same thing that I have and that doesn't match the expected behavior described in the documentation.

With `preserve`set to `true`, we have

```css
max-width: var(--content-width);
```

instead of 

```css
max-width: 45rem;
max-width: var(--content-width);
```
> @benoitchantre does setting preserve to false in config.json work for you? When I do that both the fallback and custom properties are printed.

@ataylorme I have tested changing `preserve` to `false` either in `config.json` or in `config.default.json` and see no difference in the processed stylesheets (tested with `content.min.css`) between `true` or `false` value for `preserve`.

> P.S. you can use npm run build -- styles to run just the styles task for testing.

Thank you for the tip

ataylorme:
@benoitchantre thank you for the thorough explanation. I think I have tracked down the issue.

I've been using a simpler file named `assets/css/src/custom-properties-test.css` with the contents below to test.

```css
.has-theme-red-color,
.has-theme-red-color a {
	color: var(--color-theme-red);
}
```

With `preserve` set to `true` (from `config.default.json`) the result is:

```css
.has-theme-red-color,.has-theme-red-color a{color:#c0392b;color:var(--color-theme-red)}
```

With `preserve` set to `false` in `config.json` the result is:

```css
.has-theme-red-color,.has-theme-red-color a{color:#c0392b}
```

Does this match what you expect? If so, can you test again as well?

ataylorme:
Actually, now that I test the same in `assets/css/src/editor/custom-properties-test.css` it seems that `preserve` is not being set to `false`.

I think perhaps changing the variable inline with `gulp-tap` is not working.

benoitchantre:
> Does this match what you expect? If so, can you test again as well?

Yes. It works like expected, but unfortunately not with `editor-styles.css` (the custom properties are not replaced).

ataylorme:
@benoitchantre since changing `preserve` for editor files mid-stream wasn't working I pulled in the work from #370 that uses a new task for CSS in `assets/css/src/editor`.

This is testing well for me with both editor and normal CSS. `preserve` being `true` or`false` in config effects all CSS except editor CSS, which always uses `false` for `preserve`.

benoitchantre:
@ataylorme `npm run build` fails with the latest commit.

```
[20:27:59] 'styles' errored after 1.26 s
[20:27:59] Error: write after end
    at writeAfterEnd (/Users/benoit/Git/wprig/node_modules/readable-stream/lib/_stream_writable.js:288:12)
    at DestroyableTransform.Writable.write (/Users/benoit/Git/wprig/node_modules/readable-stream/lib/_stream_writable.js:332:20)
    at DestroyableTransform.ondata (/Users/benoit/Git/wprig/node_modules/readable-stream/lib/_stream_readable.js:619:20)
    at DestroyableTransform.emit (events.js:182:13)
    at DestroyableTransform.EventEmitter.emit (domain.js:460:23)
    at addChunk (/Users/benoit/Git/wprig/node_modules/readable-stream/lib/_stream_readable.js:291:12)
    at readableAddChunk (/Users/benoit/Git/wprig/node_modules/readable-stream/lib/_stream_readable.js:278:11)
    at DestroyableTransform.Readable.push (/Users/benoit/Git/wprig/node_modules/readable-stream/lib/_stream_readable.js:245:10)
    at DestroyableTransform.Transform.push (/Users/benoit/Git/wprig/node_modules/readable-stream/lib/_stream_transform.js:148:32)
    at Pumpify.onReadable (/Users/benoit/Git/wprig/node_modules/to-through/index.js:25:14)
[20:27:59] 'buildDev' errored after 1.27 s
[20:27:59] The following tasks did not complete: php, editorStyles, scripts, translate
[20:27:59] Did you forget to signal async completion?
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! wprig@2.0.0 build: `gulp buildDev`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the wprig@2.0.0 build script.

```

ataylorme:
Oh weird - the task completes successfully in isolation with `npm run dev -- editorStyles`. I'll look into it.

ataylorme:
I think it was a race condition between `styles` and `editorStyles` writing to the same directory. I moved them to run in series instead of parallel.

Can you test again @benoitchantre?

P.S. *thank you* for all your help sorting this out and with testing.

benoitchantre:
It seems some styles are missing in `editor-styles-min.css`. I have merged this PR into #332 and found that `.wp-block`, `.wp-block[data-align="wide"]` and other styles are missing in `editor-styles.min.css`.

I have also found that turning `debug` to `true` doesn't change the output for `editor-styles.min.css`: it's always minified.

> P.S. thank you for all your help sorting this out and with testing.

You're welcome. Thank you a lot for your work on the build process.

ataylorme:
Hmm I am not seeing those same results.

I updated `assets/src/editor/edtior-styles.min.css` to the below and set `debug` to `true`.

```css
.has-theme-red-color,
.has-theme-red-color a {
	color: var(--color-theme-red);
}
.wp-block[data-align="wide"] {
	max-width: calc(50% + var(--content-width) / 2 + 76px);
}
```

This was the result

```css
.has-theme-red-color,
.has-theme-red-color a {
	color: #c0392b;
}

.wp-block[data-align="wide"] {
	max-width: calc(50% + 45rem / 2 + 76px);
}
```

benoitchantre:
I get the expected result with `npm run dev -- editorStyles`, but `npm run dev` and `npm run build` don't create `editor-styles.min.css`.

Why do we have `dev` in the command to process the editor styles (`npm run dev -- editorStyles`)? `npm run dev` has a watch task. I think either `npm run dev -- editorStyles` should have a watch task or it should be renamed `npm run build -- editorStyles`.

Good news: after a merge into #332, the styles are inlined properly.

ataylorme:
> but `npm run dev` and `npm run build` don't create `editor-styles.min.css`.

You were right. I didn't catch it since those files already existed for me.

I added some clean tasks so now on build/at start-up existing minified CSS and JS files are deleted first.

I think this is fixed now.

> Why do we have `dev` in the command to process the editor styles (`npm run dev -- editorStyles`)? 

We are using [NPM run-script](https://docs.npmjs.com/cli/run-script), which allows you to specify command aliases in `package.json`. We are using this because it also uses the binaries in `node_modules`, this way the version of `gulp` run is from NPM and not on the user's machine.

`npm run dev` is aliased to `gulp`. `gulp`, without as task specified, runs the default task, which in our case is the dev/watch task.

However, using a double dash (`--`) you can pass extra parameter to `npm run` commands. So, `npm run dev -- editorStyle` is really `gulp editorStyle`, thus running the `editorStyle` task.

I guess we could add a `gulp` alias. `npm run gulp -- <taskName>` might be less confusing.

benoitchantre:
> I think this is fixed now.

Yes, all commands produce the same result.

Unfortunately, the result is not what we expect:

```css
.block-editor-page .editor-styles-wrapper {
	color: #333;
	color: #333;
	color: var(--global-font-color);
	font-family: "Crimson Text", serif;
	font-family: "Crimson Text", serif;
	font-family: var(--global-font-family);
	/* stylelint-disable */
	font-size: calc(20 / 16 * 1rem);
	font-size: calc(20 / 16 * 1rem);
	font-size: var(--font-size-regular);
	/* stylelint-enable */
	line-height: 1.4;
	line-height: 1.4;
	line-height: var(--global-font-line-height);
}
```

instead of 

```css
.block-editor-page .editor-styles-wrapper {
	color: #333;
	font-family: "Crimson Text", serif;
	/* stylelint-disable */
	font-size: calc(20 / 16 * 1rem);
	/* stylelint-enable */
	line-height: 1.4;
}
```
> I guess we could add a gulp alias. npm run gulp -- <taskName> might be less confusing.

I think its a good idea. Thank you for the explanation.

ataylorme:
```css
color: #333;
color: #333;
color: var(--global-font-color);
```

@benoitchantre I think this is because of the `@import "../custom-media.css";` and `@import "../custom-properties.css";` in the editor source CSS in combination with the [`importFrom` option of postcss](https://github.com/postcss/postcss-custom-properties#importfrom).

The variable are being imported in the build process via `importFrom` and also imported in the CSS file and processed with `postcss-import`.

One of these should probably be removed. The options are:
* Keep importing CSS using `@import` statements. Remove `importFrom` from the config and build process.
* Keep `importFrom` from the config and build process. Remove importing CSS using `@import` statements

@mor10 might have an opinion. What are your thoughts?

Personally, I like removing `@import`. Defining CSS property and media file in config and then having them injected with `importFrom` seems cleaner than asking users to remember when/where they need to use `@import`, especially as they add/delete CSS files.

ataylorme:
> I guess we could add a gulp alias. npm run gulp -- might be less confusing.
>>I think its a good idea. Thank you for the explanation.

I added this. You can now use `npm run gulp` and pass in any defined task. For example, `npm run gulp -- editorStyles`.

benoitchantre:
When I remove `@import`, here's what we get.
It looks like the behaviour when `preset` is set to `true`

```css
.block-editor-page .editor-styles-wrapper {
	color: #333;
	color: var(--global-font-color);
	font-family: "Crimson Text", serif;
	font-family: var(--global-font-family);
	/* stylelint-disable */
	font-size: calc(20 / 16 * 1rem);
	font-size: var(--font-size-regular);
	/* stylelint-enable */
	line-height: 1.4;
	line-height: var(--global-font-line-height);
}
```

When I remove `importFrom` in `editorStyles.js`, here's what we get:
```css
.block-editor-page .editor-styles-wrapper {
	color: #333;
	color: #333;
	color: var(--global-font-color);
	font-family: "Crimson Text", serif;
	font-family: "Crimson Text", serif;
	font-family: var(--global-font-family);
	/* stylelint-disable */
	font-size: calc(20 / 16 * 1rem);
	font-size: calc(20 / 16 * 1rem);
	font-size: var(--font-size-regular);
	/* stylelint-enable */
	line-height: 1.4;
	line-height: 1.4;
	line-height: var(--global-font-line-height);
}
```

Expected output:


```css
.block-editor-page .editor-styles-wrapper {
	color: #333;
	font-family: "Crimson Text", serif;
	/* stylelint-disable */
	font-size: calc(20 / 16 * 1rem);
	/* stylelint-enable */
	line-height: 1.4;
}
```

I need to stop now. I'll continue the tests on Monday if necessary.
Thank you again for your work.

ataylorme:
> 

Understood, enjoy your weekend 😄.

I'm on the road and not doing my best work either - comitting between conference session. I forgot to force `preserve` to `false` for `editorStyles` when refactoring and breaking it out into its own task.

[`2943887`](https://github.com/wprig/wprig/pull/369/commits/29438875cf6207ece5bda80e188ea8e64bdb84d3) should have fixed that.

The good news is I am no longer seeing double imports in editor styles, but they persist in `assets/css/global.min.css`, which has `@import`. `vim assets/css/content.min.css`, which doesn't use `@import`, does not have double declarations

benoitchantre:
@ataylorme `editor-styles.min.css` contains the expected result 😄  but `global.min.css` has double imports as you mentioned.

If we remove `@import` in favor of `importFrom`, I think we will increase the complexity for theme authors in the case they need to handle imports differently. We might want to split `content.css` and `global.css` into different parts to import some of them in `editor-styles.css` to keep things DRY.

ataylorme:
> If we remove @import in favor of importFrom, I think we will increase the complexity for theme authors in the case they need to handle imports differently. We might want to split content.css and global.css into different parts to import some of them in editor-styles.css to keep things DRY.

It sounds like you are in favor of continuing to import CSS using `@import` statements and removing `importFrom` from the config and build process.

I can move ahead with that if Morten is in agreement.

> We might want to split `content.css` and `global.css` into different parts to import some of them in `editor-styles.css` to keep things DRY.

Could this be done in #332 or a follow-up PR? This one has gotten quite large as it is.

benoitchantre:
> Could this be done in #332 or a follow-up PR?

I'll work on that in a follow-up PR. That will allow us tomove forward by merging this one and #332, and discuss enhancements separately.

mor10:
My vote is to use `@import` and remove `importFrom`. `@import` within the original CSS files is easier to understand and manage because it happens in context. In my opinion.

ataylorme:
Since we are all on the same page about removing `importFrom` I've done that.

@mor10 and @benoitchantre please review + test one more time, then we can (hopefully) merge this!

benoitchantre:
Thank you for the latest changes @ataylorme. I just tested it and still see double imports in `global.css.min`

Here's an example:
```css
a {
	color: #0073aa;
	color: #0073aa;
	color: var(--color-link);
}
```

benoitchantre:
It's caused by`postcss-preset-env` which adds fallbacks, like `postcss-custom-properties` and `postcss-custom-media`.

ataylorme:
Bah, sorry about that @benoitchantre - I forgot to `git add` the `gulp/editorSyles.js` file so the `importFrom` removal there was not committed. It should be now.

ataylorme:
> It's caused by `postcss-preset-env` which adds fallbacks, like `postcss-custom-properties` and `postcss-custom-media`.

Ah, I see it is a different problem.

benoitchantre:
One more thing: shouldn't we move the imported files into a directory? It would be easier to understand what happens by looking at how the files are organized and we could ignore these files based on the directory name instead of an array of filenames.

ataylorme:
It looks like `postcss-preset-env` takes care of custom properties and custom media, so `postcss-custom-properties` and `postcss-custom-media` might not be needed.

I've removed it and see the duplications gone in `assets/css/global.min.css`.  `assets/css/editor/editor-styles.min.css` still looks okay too.

`assets/css/content.min.css` is missing the fallbacks since it does not `@import` the custom properties in the file. However, the CSS still works in Chrome for me as the custom properties are read from `global.css`, which is loaded first.

ataylorme:
> shouldn't we move the imported files into a directory? It would be easier to understand what happens by looking at how the files are organized and we could ignore these files based on the directory name instead of an array of filenames.

Let's do that in a follow up PR as well.

benoitchantre:
> `assets/css/content.min.css` is missing the fallbacks since it does not `@import` the custom properties in the file. However, the CSS still works in Chrome for me as the custom properties are read from `global.css`, which is loaded first.

Works also in Firefox, but a browser that doesn't understand custom properties will have troubles with declarations that have no fallback. A simple fix is to add `@import "custom-media.css";` and
 `@import "custom-properties.css";` to all css files.

ataylorme:
> A simple fix is to add @import "custom-media.css"; and
@import "custom-properties.css"; to all css files.

If we are going to import on every file *and* move to a dedicated directory for imports, I think going back to `importFrom` makes sense to me since changing what gets imported would mean editing the config/files in the import directory instead of editing the include statements in every CSS file.

We could also leave it as-is and just document that folks will need to `@import` to all files if they need legacy browser support.

benoitchantre:
If we use `importFrom`, how can we stay flexible without adding to much complexity if we need other imports for specific files? Exemple of use case: many styles from `global.css` could be move to a separated file to be imported in `global.css` and `editor-styles.css`.

ataylorme:
> One more thing: shouldn't we move the imported files into a directory?

I thought you meant only the custom property and custom media files. For example, if we had a new directory `assets/css/src/imports` that was defined in the configuration we could use `importFrom` to add them to all processed CSS automatically, as opposed to using `@import` in all CSS files for bringing them in.

> how can we stay flexible without adding to much complexity if we need other imports for specific files? Exemple of use case: many styles from global.css could be move to a separated file to be imported in global.css and editor-styles.css.

I think using `@import` in this case would be fine

benoitchantre:
In `editorStyles.js`, we must force `preserve` to `false`, otherwise styles are broken in the editor because of undefined `var`.

Changes to `editorStyles.js` in https://github.com/wprig/wprig/pull/369/commits/d9a63bdeda920a214d0747dfbca0e234548fff62 should be reverted.

ataylorme:
@benoitchantre sorry, I got pulled into a work meeting. `postcss-preset-env` supports `preserve`](https://www.npmjs.com/package/postcss-preset-env#preserve) so I moved the `preserve` `false` declaration there, rather than adding those packages back.

benoitchantre:
By removing `postcssCustomProperties` and `postcssCustomMedia` from `styles.js` https://github.com/wprig/wprig/commit/d9a63bdeda920a214d0747dfbca0e234548fff62, the `preset` value in the config file is not used. We should respect the value from the config file.

> sorry, I got pulled into a work meeting

No problem. On my side, it's already late. I'll continue tomorrow.

ataylorme:
> By removing postcssCustomProperties and postcssCustomMedia from styles.js in d9a63bd, the preset value from the config file is not used. We should respect the value from the config file.

Agreed, I added that back in [`0ad26c1`](https://github.com/wprig/wprig/pull/369/commits/0ad26c114be22f3b5868f7b28d1977383da415f3).

> On my side, it's already late. I'll continue tomorrow.

Have a great evening! 👋

benoitchantre:
I have noticed some visual changes, due to custom media queries that are not computed.

For example: `@media (--sidebar-query) {` is left as is instead of beeing computed to `@media screen and (min-width: 60em) {`

Custom media queries is in `stage 1` from [`postcss-preset-env` features documentation](https://preset-env.cssdb.org/features). `stage 3` is used by WP Rig build process. Changing `stage 3` to `stage 1` fixes this regression.

ataylorme:
Thanks @benoitchantre. Rather than switch to stage `1` I used [the `features` options of `postcss-preset-env`](https://www.npmjs.com/package/postcss-preset-env#features), which allows you to enable specific polyfills regardless of whatever stage is defined. I also added `stage` to `config.default.json` with a default of `3`.

> For example: `@media (--sidebar-query) {` is left as is instead of beeing computed to `@media screen and (min-width: 60em) {`

When testing `front-page.css` I see this behavior _unless_ I add `@import "./custom-media.css"; @import "./custom-properties.css";` to the top of the file, which I think is to be expected.

This is the same way custom properties are acting in our previous observation now that we are not injecting custom media and custom properties into all files with `importFrom`.

> `assets/css/content.min.css` is missing the fallbacks since it does not `@import` the custom properties in the file. However, the CSS still works in Chrome for me as the custom properties are read from `global.css`, which is loaded first.

benoitchantre:
> When testing front-page.css I see this behavior unless I add `@import "./custom-media.css"`; `@import "./custom-properties.css"`; to the top of the file, which I think is to be expected.

@ataylorme Can you import `custom-media.css` and `custom-properties.css` in all files with `@import` (KISS)?

I'll do a test after that step, which should be the last commit 🤞

ataylorme:
Yes, but it will be a few hours. I am walking to the gym for a morning workout at the moment and not at the computer.

benoitchantre:
> Yes, but it will be a few hours. I am walking to the gym for a morning workout at the moment and not at the computer.

No problem. 🏋️

ataylorme:
@benoitchantre I made a few changes.

First, I simplified the ignored rules so that all CSS/JS files with a `_` prefix are not built/minified. This made the path configuration in gulp much easier and also made it so that when new partial files are created they do not need to be added to the config.

After that, I added `@import "_custom-properties.css";` and `@import "_custom-media.css";` to the top of all the CSS source files.

I also added the opt-in features (custom media, custom properties, and nesting) for `postcss-preset-env` to `config.default.json`. `stage` and `preserve` are there as well.

I tested `npm run build` and `npm run dev` and things are working well for me.

Overall, I think this entire process left us in a better state. The ignores are not a mess and we are using `postcss-preset-env`, which calls the other `postcss` plugins for us under the hood, simplifying a lot of config.

ataylorme:
I updated the PR description to reflect all the changes. @mor10 can you give this a spin?

benoitchantre:
Thank you @ataylorme. I should find the time to look at that in 3-4 hours.

benoitchantre:
It works well @ataylorme. Thank you for your patience and efforts on this.

@mor10 Could you explain why we should preserve custom properties and custom media? It is to ease tweaks with the inspector during the development phase? Custom media with fallbacks duplicate many declarations (see the example below from `global.min.css`).

```css
@media screen and (min-width: 60em) {

	.has-sidebar .site {
		margin: 0 1em;
		display: grid;
		grid-template-columns: 3fr 1fr;
		grid-gap: 2rem;
		-webkit-box-pack: center;
		-ms-flex-pack: center;
		justify-content: center;
	}

	.site-header {
		grid-column: span 2;
	}

	.site-main {
		grid-column: 1;
		min-width: 0;
	}

	.primary-sidebar {
		grid-column: 2;
		min-width: 0;
	}

	.site-footer {
		grid-column: span 2;
	}
}

@media (--sidebar-query) {

	.has-sidebar .site {
		margin: 0 1em;
		display: grid;
		grid-template-columns: 3fr 1fr;
		grid-gap: 2rem;
		-webkit-box-pack: center;
		-ms-flex-pack: center;
		justify-content: center;
	}

	.site-header {
		grid-column: span 2;
	}

	.site-main {
		grid-column: 1;
		min-width: 0;
	}

	.primary-sidebar {
		grid-column: 2;
		min-width: 0;
	}

	.site-footer {
		grid-column: span 2;
	}
}
```

Shouldn't we force `preserve` to `false` for the `bundle` task in order keep production files as small as possible?

ataylorme:
> Shouldn't we force `preserve` to `false` for the bundle task in order keep production files as small as possible?

That is a good observation. I can implement this easily if it is decided we should do that.

ataylorme:
@benoitchantre Morten and I discussed the best way to solve for older browsers in WP Rig office hours today.

tl;dr the default is using custom properties and custom variables as-is (modern browser support) with injection optional if you need legacy browser support.

The long explanation is below.

`preserve` is no longer configurable. Instead, there is a new configuration option `dev.styles.importFrom` that takes an array of CSS files, relative to the CSS source path, to be injected at build time.

Here is an example `config.json`:

```json
{
  "dev": {
    "styles": {
      "importFrom": ["_custom-properties.css", "_custom-media.css"]
    }
  }
}
```

When custom media and custom property files are injected like this, `preserve` is automatically set to `false`.

If the list of files to inject is empty, then `preserve` is `true` (except the editor styles, which are hard-coded to `false`).

If custom media and custom property files are not being injected, then the assumption is that they will be pulled in using `@import` in the individual source CSS files.

ataylorme:
@mor10 this is ready for review again. It turns out custom media *do* need to be imported to every file. The custom media declarations do not have a global scope, like custom properties in `:root`, and are bound to the individual CSS file.

mor10:
> When custom media and custom property files are injected like this, `preserve` is automatically set to `false`.

Is it possible to make it so this creates the _fallback_ rather than an out and out replacement the way it was earlier? I'd like for the custom properties to be preserved in production no matter what people do.

ataylorme:
> Is it possible to make it so this creates the fallback rather than an out and out replacement the way it was earlier? I'd like for the custom properties to be preserved in production no matter what people do.

I changed this because if you inject custom properties `importFrom` and do *not* import them with `@import` in the source CSS files then it causes issues because the variable declarations are printed after the fallback (example below) but they are not defined.

```
.alignwide {
    max-width: calc(50% + 45rem / 2);
    max-width: calc(50% + var(--content-width) / 2);
}
```

`var(--content-width)` is undefined and declared second so things fall apart.

ataylorme:
@mor10 can you test with the following:

* `@import "_custom-properties.css";` removed from `assets/css/src/global.css`
* `preserve` hard-coded to `true` in `gulp/styles.js` (line 63)
* set `"importFrom": ["_custom-properties.css", "_custom-media.css"]` in `config.json`

mor10:
Because the original custom properties are parsed and removed in `global.css`?

ataylorme:
If you are using `importFrom` to process custom properties at build time then `@import` isn't needed in the source CSS files.

If that is done and `preserve` is `true`, leaving things like `var(--color-quote-citation)` in the outputed CSS.

That is not good since `--color-quote-citation` is undefined (the `@import` into `global.css` to define the CSS vars in `:root` is gone).

mor10:
The above request looks like it's been resolved. On testing I see the custom property declarations in `global.min.css` and the custom properties with fallbacks in the other files.

benoitchantre:
> @benoitchantre Morten and I discussed the best way to solve for older browsers in WP Rig office hours today.

Thank you @ataylorme for the summary.

@mor10 @ataylorme What did you decide about production files? Custom media with their fallback duplicate many declarations. At the moment, I don't understand why we should keep them for production files?

ataylorme:
> What did you decide about production files?

We did not discuss what should be done in production

> Custom media with their fallback duplicate many declarations. At the moment, I don't understand why we should keep them for production files?

I just pushed up a change that sets `preserve` to `false` for production.

@benoitchantre I'm interested to see if the `importFrom` config option works for your use case.

ataylorme:
> I just pushed up a change that sets preserve to false for production.

@mor10 says take it back. I will let him elaborate.

mor10:
> Shouldn't we force `preserve` to `false` for the `bundle` task in order keep production files as small as possible?

@benoitchantre Take a look at the updated PR and how things work now. While I agree the doubling up of rules based on media queries is ... not great, it is a temporary problem as browsers get up to standard. I'd rather we ship code marked up to modern standards with fallbacks than ship code which holds us back.

ataylorme:
> What did you decide about production files? Custom media with their fallback duplicate many declarations. At the moment, I don't understand why we should keep them for production files?

I think we should decide this is a follow-up issue. It will be a one-line change for an easy follow on PR if one is needed.

For now, I would like to see this merged.

benoitchantre:
I have noticed that fallbacks are only in `global.min.css`. They miss in other files.

> I'm interested to see if the `importFrom` config option works for your use case.

When I use the `importFrom` option, I see custom properies and fallback in all files. From your explanation, I should only have the computed values.

No issue with editor styles in combination with #332.

ataylorme:
> When I use the importFrom option, I see custom properties and fallback in all files. From your explanation, I should only have the computed values.

That changed when `preserve` got hard-coded to `true` in [`b4b501b`](https://github.com/wprig/wprig/pull/369/commits/b4b501b22f989817319b0f4348a019f463a03a55)

See @mor10 comment above

> While I agree the doubling up of rules based on media queries is ... not great, it is a temporary problem as browsers get up to standard. I'd rather we ship code marked up to modern standards with fallbacks than ship code which holds us back.

@mor10 do you object to `preserve` being `false` even when `importFrom` is used?

benoitchantre:
> That changed when `preserve` got hard-coded to `true` in https://github.com/wprig/wprig/pull/369/commits/b4b501b22f989817319b0f4348a019f463a03a55

It doesn't work as it should: with `preserve` set to `true`, we should have custom properties *and* fallbacks *in all files*. I only see fallbacks in `global.min.css`.

ataylorme:
>  I only see fallbacks in global.min.css.

This is because custom properties are only included via `@import` in `assets/css/src/global.css`.

In browsers that support custom properties the inclusion in `global.css` declares the variables in the `:root` scope and the other CSS files can use the variables without needing to redeclare them or have the actual fallback values printed.

> we should have custom properties and fallbacks in all files.

This is needed for browsers that do not support custom properties and is what the optional `importFrom` injection is for. As @mor10 said above we don't want that to be the default as it is a workaround until browser support for custom properties catches up.

benoitchantre:
> In browsers that support custom properties

It works fine for them. The issue is for browsers that don't support custom properties, because need fallbacks to render the page properly. I'm not saying that we should change the `preserve` value, only that using `true` should produce fallbacks and keep custom properties in order to be compatible with all browsers.

ataylorme:
> The issue is for browsers that don't support custom properties, because need fallbacks to render the page properly.

Indeed, that is an issue if your project will be built for browsers that do not support custom properties.

I think what @mor10 was trying to say above is that we want to assume browsers support custom properties so that as support increases we don't have unnecessary fallbacks.

Therefore, the default is to process the build as if browsers support custom properties.

>  I'm not saying that we should change the preserve value, only that using true should produce fallbacks and keep custom properties in order to be compatible with all browsers.

This will only happen if the custom property declarations are available to each file at build time. Right now, the custom property definitions are only `@import`ed in `global.css`. Doing a direct `@import` of custom properties in all CSS source files would cause the built CSS to have multiple declarations for each custom property (once in each file).

Instead, the `importFrom` option is available in the configuration to inject the custom properties to each file at build time for projects that need the fallback in each file.

Since you personally want fallbacks, are you okay using the `importFrom` config override in your projects (rather than having fallbacks be the default)?

benoitchantre:
Thank you for the detailed explanation @ataylorme, I have now a better understanding of what was intended.

> Since you personally want fallbacks, are you okay using the `importFrom` config override in your projects (rather than having fallbacks be the default)?

Yes

ataylorme:
> Thank you for the detailed explanation @ataylorme, I have now a better understanding of what was intended.

You are welcome! Thank you for sticking with us through sorting this all out.

With your updated understanding of the intention, do things work as you expect?

benoitchantre:
> With your updated understanding of the intention, do things work as you expect?

Yes 👏 

In `CHANGLOG.md`, lines 30-31, the path shoud be updated to include the underscore before `custom-properties.css` and `custom-media.css`

ataylorme:
> In CHANGLOG.md, lines 30-31, the path shoud be updated to include the underscore before custom-properties.css and custom-media.css

Done!

mor10:
Looks like we're all in agreement. Now we need one more core contributor to sign off and get this ball rolling! Paging @bamadesigner

Issue #391, opened 2019-03-29
Add wp_body_open hook


benoitchantre:
## Issue Overview

> Starting from WordPress 5.2 a new function is added – `wp-body_open()` – that is used to trigger a `wp_body_open` action. The intention of this action is to allow people to add things – like a `<script>` tag – directly inside the body of a page.
> Themes should start adding this hook to their themes as soon as 5.2 releases.

Source : [make.wordpress.org](https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/)

bamadesigner:
Addressed in #392.

PR #392, opened 2019-03-29
Add action wp body open

benoitchantre:
## Description

Addresses issue #391.

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/header.php
+++ b/header.php
@@ -31,6 +31,7 @@ namespace WP_Rig\WP_Rig;
 </head>
 
 <body <?php body_class(); ?>>
+<?php do_action( 'wp_body_open' ); ?>
 <div id="page" class="site">
 	<a class="skip-link screen-reader-text" href="#primary"><?php esc_html_e( 'Skip to content', 'wp-rig' ); ?></a>
 


benoitchantre:
@felixarntz Is it okay for you to use directly `do_action( 'wp_body_open' );` in `header.php` or would you prefer to use a shim in *Base_Support* component

```php
if ( ! function_exists( 'wp_body_open' ) ) {
        function wp_body_open() {
                do_action( 'wp_body_open' );
        }
}
```

ataylorme:
@benoitchantre I think we should add the shim. It won't hurt anything for folks on WordPress `5.2` or greater and will keep things from breaking for those on a lower version.

header.php
```diff
@@ -31,6 +31,7 @@
 </head>
 
 <body <?php body_class(); ?>>
+<?php do_action( 'wp_body_open' ); ?>
```
Ismail-elkorchi: `do_action( 'wp_body_open' )` is already defined in [/wp-includes/general-template.php](https://github.com/WordPress/WordPress/blob/master/wp-includes/general-template.php#L2765). To trigger the `wp_body_open` action, one can call simply the `wp_body_open()` function.

benoitchantre: This function simply triggers `do_action( 'wp_body_open' )`. Technically, we can call `do_action( 'wp_body_open' )` directly. It's only a matter of style

benoitchantre:
> I think we should add the shim. It won't hurt anything for folks on WordPress 5.2 or greater and will keep things from breaking for those on a lower version.

@ataylorme Nothing break if we trigger an undefined action: it simply does nothing.

If we go with the shim, I haven't found something better than putting it in `function.php`, right after the constants.

ataylorme:
> Nothing break if we trigger an undefined action: it simply does nothing.

Okay, I am fine with omitting it then

header.php
Ismail-elkorchi: `wp_head()` and `wp_footer()` are also wrappers for `do_action( 'wp_head' )` and `do_action( 'wp_footer' )`, and yet, using these directly is not recommended.

mor10:
I'm in agreement with @bamadesigner: Shim required. This is too new and too failure-prone atm.

ataylorme:
> Ideally, I would like to have the shim in `back-compat.php`, but it involve to rethink `back-compat.php`.

Can we move the shim to `back-compat.php` now and do a refactor of `back-compat.php` in a follow up PR?

ataylorme:
In office hours @benoitchantre and I discussed making a new file `back-compat-wordpress.php` for WordPress shims. We can rename the current `back-compat.php` to back-compat-php.php`. Both will be loaded by `functions.php`.

benoitchantre:
@ataylorme `back-compat.php` actually contains test for WordPress and PHP versions, then prevent the user to switch to WP Rig. I think it would be more straightforward to keep `back-compat.php` untouched and add `wordpress-shims.php`. I think it was your first idea.

```php
// Bail if requirements are not met.
if ( version_compare( $GLOBALS['wp_version'], WP_RIG_MINIMUM_WP_VERSION, '<' ) || version_compare( phpversion(), WP_RIG_MINIMUM_PHP_VERSION, '<' ) ) {
	require get_template_directory() . '/inc/back-compat.php';
	return;
}

// Include WordPress shims
require get_template_directory() . '/inc/wordpress-shims.php';
```

What do you think?

ataylorme:
@benoitchantre that works for me - thanks for clarifying!

ataylorme:
> Indent with tabs

I was trying to clean up code sniffing errors and only managed to introduce more 🤦‍♂️

bamadesigner:
Thanks @benoitchantre!

Issue #371, opened 2019-03-09
Add tests for gulp tasks


ataylorme:
## Issue Overview
As the gulp build process is evolving, manual testing has become cumbersome. We should automate testing of gulp tasks to avoid introducing regressions.

https://snugug.com/musings/unit-testing-gulp-tasks/ and https://github.com/avajs/ava look promising.

ataylorme:
Closed by #429

PR #397, opened 2019-03-31
Single stream gulp tasks

ataylorme:
## Description
Use [`pipeline` from `mississippi`](https://www.npmjs.com/package/mississippi#pipeline) to convert the tasks in before replacement and after replacement into a single stream. This allows a much cleaner use of `pump`.

The tasks to run before and after replacement are now in exported functions so that they can be re-used. This is a pre-requisite for writing tests for gulp tasks (#371).

Props @phated!

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/gulp/styles.js
+++ b/gulp/styles.js
@@ -102,16 +102,13 @@ export function stylesAfterReplacementStream() {
 */
 export default function styles(done) {
 
-	// Convert the array of string replacement tasks into a single stream object
-	const stringReplacementTasks = pipeline.obj( getStringReplacementTasks() );
-
-	pump([
+	return pump([
 		src( paths.styles.src, {sourcemaps: !isProd} ),
 		stylesBeforeReplacementStream(),
 		// Only do string replacements when building for production
 		gulpPlugins.if(
 			isProd,
-			stringReplacementTasks
+			getStringReplacementTasks()
 		),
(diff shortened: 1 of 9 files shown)


ataylorme:
@phated mind doing a sanity check on the updates to the `gulp/styles.js` task here?

If it looks good I plan to make similar adjustments to the other tasks. Once that is done, then I can make a follow-up PR with the actual tests.

gulp/styles.js
```diff
@@ -53,56 +71,49 @@ export default function styles(done) {
...
-				[]
-			),
-			afterReplacement
+/**
+* CSS via PostCSS + CSSNext (includes Autoprefixer by default).
+*/
+export default function styles(done) {
+
+	// Convert the array of string replacement tasks into a single stream object
+	const stringReplacementTasks = pipeline.obj( getStringReplacementTasks() );
```
phated: I'd probably move this into the `getStringReplacementTasks()` method.

ataylorme: Yeah, that’s a good suggestion. Right now some of the other tasks expect an array.

I’ll update the rest of the tasks to use pipeline then update the return type of getStringReplacementTasks

ataylorme:
> Works for me.

@mor10 did you test the gulp processes?

Issue #403, opened 2019-04-02
Remove @custom-media rules from setup (for now)


mor10:
See #400 and https://github.com/wprig/wprig/pull/369#issuecomment-472177337

`@custom-media` might be a bridge too far along the bleeding edge route for right now. I propose we remove the `@custom-media` rules from existing stylesheets, replacing them with standard Custom Properties for now, as we await better support. The custom media rules file can stay in the system with a comment at the top saying it is experimental and conflicts with AMP plugin etc. 

This resolves #400 without changing the Gulp + PostCSS configuration, leaves developers to play around with bleeding edge experimental features if they want to, and provides necessary context for people to make informed decisions.

CC @ataylorme @westonruter @benoitchantre

Tagging as `v2.0` bug because it is.

ataylorme:
I am okay with removing custom media but leaving support for it (and documenting that support) if user's wish to implement it. We could warn about possible AMP incompatibilities.

To me, this is similar to WP Rig supporting Sass but not shipping with any Sass files.

mor10:
Riiiiiiiiiight... There was a reason I used `@custom-media` to begin with: You can't use custom properties in media queries, which is why `@custom-media` exists in the first place. This requires some more thought.

From the [spec](https://www.w3.org/TR/css-variables-1/#using-variables):

> The var() function can not be used as property names, selectors, or anything else besides property values. (Doing so usually produces invalid syntax, or else a value whose meaning has no connection to the variable.)

Thoughts @bamadesigner and @benoitchantre?

benoitchantre:
Ok with your proposal.

Would it be possible to check and merge #390 before?

ataylorme:
> The use of @custom-media is not allowed with AMP.

Custom media just needs to be excluded from the built CSS, not the source. We could rethink the decision in #369 and use `importFrom` to inject/process custom-media/custom-properties and _not_ preserve them in the built CSS.

This would allow them to continue to be used in the source without having anything in the built CSS that conflicts with AMP.

mor10:
Custom Properties must be preserved. They are not bleeding edge. Custom media can be imported and rewritten in the interim.

ataylorme:
Closed by #405

Issue #400, opened 2019-04-01
AMP error for @custom-media


teddarling:
## Issue Overview
The use of @custom-media is not allowed with AMP.

## Describe your environment
Using VS Code, Chrome/Safari/Firefox.

## Steps to Reproduce
1. Build WP Rig
2. Install WP Rig
3. Install AMP Plugin (https://wordpress.org/plugins/amp/)Install AMP Plugin (https://wordpress.org/plugins/amp/)
4. Activate AMP Plugin
5. Validate any page
6. View AMP Validation Errors
7. Notice result.

## Expected Behavior
A valid amp page on install of AMP and Built WP Rig setup.

## Current Behavior
AMP provides the following error:
- illegal_css_at_rule: @custom-media

## Possible Solution
Use either hard coded CSS media query values or use SASS to build the CSS with SASS variables.

## Screenshots / Video
![AMP Error](https://user-images.githubusercontent.com/3044359/55365120-82f36300-54a0-11e9-9518-877c984f0cfc.png)

GitaStreet:
Yes, this issue also makes variables disappear from:
 _custom-properties.css

Commented this line out in various .css files - and the variable definitions are back:
@import "_custom-media.css

GitaStreet:
Found this suggestion that seems to work:
"CSS variables should to be set in :root { }"
vuejs-templates/webpack#413

Quickfix not thoroughly tested:

Uncomment or delete all instances of:
@import "_custom-media.css";
Move these lines of code into the :root{} in the file _custom-properties.css
/* stylelint-disable */ @custom-media --narrow-menu-query screen and (max-width: 37.5em); @custom-media --wide-menu-query screen and (min-width: 37.5em); @custom-media --content-query screen and (min-width: 48em); @custom-media --sidebar-query screen and (min-width: 60em); /* stylelint-enable */

Error message still there, but theme appears to render correctly.

teddarling:
First, Let me add I updated my initial post to indicate that this is in version 2.0

@GitaStreet, I did add the variables into custom-properties.css, as you mentioned.

```
	/* stylelint-disable */
	@custom-media --narrow-menu-query screen and (max-width: 37.5em);
	@custom-media --wide-menu-query screen and (min-width: 37.5em);
	@custom-media --content-query screen and (min-width: 48em);
	@custom-media --sidebar-query screen and (min-width: 60em);
	/* stylelint-enable */
```

[...]

PR #390, opened 2019-03-28
Reuse styles in the editor

benoitchantre:
## Description

Addresses issue #334.

## List of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] This pull request relates to a ticket.
- [x] This code is tested.
- [ ] This change has been added to CHANGELOG.md
- [x] I want my code added to WP Rig.

PR #405, opened 2019-04-02
Set preserve to false for custom-media

ataylorme:
## Description

Addresses issues #403 and #400.

`fd1f88a` sets postCSS `preserve` to `false` for `custom-media`, while leaving it `true` for `custom-properties`.

`8eb460c` is optional and globally imports `custom-media` in configuration rather than each individual CSS file.

## List of changes

## Checklist:
- [x] This code is tested.


gulp/styles.js
```diff
@@ -61,13 +61,16 @@ export function stylesAfterReplacementStream() {
 				config.dev.styles.stage :
 				3
 			),
-			preserve: true,
```
ataylorme: Removing the global `preserve: true` is necessary as setting it overrides any individual options passed.

gulp/styles.js
```diff
@@ -61,13 +61,16 @@ export function stylesAfterReplacementStream() {
...
 				configValueDefined('config.dev.styles.features') ?
 				config.dev.styles.features :
 				{
-					'custom-media-queries': true,
-					'custom-properties': true,
+					'custom-media-queries': {
+						preserve: false
+					},
+					'custom-properties': {
+						preserve: true
```
ataylorme: Leave `preserve` for `custom-properties` set to `true`.

gulp/styles.js
```diff
@@ -61,13 +61,16 @@ export function stylesAfterReplacementStream() {
...
 			),
-			preserve: true,
 			features: (
 				configValueDefined('config.dev.styles.features') ?
 				config.dev.styles.features :
 				{
-					'custom-media-queries': true,
-					'custom-properties': true,
+					'custom-media-queries': {
+						preserve: false
```
ataylorme: Set `preserve` for `custom-media` set to `false`. This will stop `custom-media` like `@media (--content-query) {` from printing in additon to the computer output (e.g. `@media screen and (min-width: 48em) {`), making the built CSS AMP compatible.

config/config.default.json
```diff
@@ -14,7 +14,9 @@
     },
     "styles": {
       "stage": 3,
-      "importFrom": [],
+      "importFrom": [
+        "_custom-media.css"
```
ataylorme: Unlike `custom-properties`, which can be declared once in `:root` of a single CSS file, `custom-media` must be declared in each CSS file. Instead of having to do `@import` for `custom-media` in each CSS file it makes sense to use `importFrom` to import `custom-media` into all CSS files at build time.

benoitchantre:
Works for me.

mor10:
Will test tonight

mor10:
Some updates are needed after the merge of #390. Sorry 😬

ataylorme:
@mor10 should be okay now - there was only one small merge conflict

benoitchantre:
Works for me.

teddarling:
Working here for me

Issue #407, opened 2019-04-02
Duplicated selectors are not found when they come from an import


benoitchantre:
## Issue Overview

Duplicate selectors and other errors are not detected because style lint is done before imports.

benoitchantre:
Cc. @ataylorme for feedback.

ataylorme:
I can't reproduce. [`assets/css/src/content.css` in the current commit on `v2.0`](https://github.com/wprig/wprig/blob/425552c5058351e047e4bc0a138c7705bd38bb1f/assets/css/src/content.css) has a duplicate `.alignwide` declaration and I am seeing it reported.



```sh
npm run gulp -- styles                                                                                                               04/2/19 at 2:00:06

> wprig@2.0.0 gulp /Users/andrewtaylor/Sites/ataylorme-wordpress/web/wp-content/themes/wprig
> gulp "styles"

[14:00:26] Requiring external module @babel/register
[14:00:29] Using gulpfile ~/Sites/ataylorme-wordpress/web/wp-content/themes/wprig/gulpfile.babel.js
[14:00:29] Starting 'styles'...

assets/css/src/content.css
264:1	⚠  Unexpected duplicate selector ".alignwide", first used at line 176 (no-duplicate-selectors) [stylelint]
268:1	⚠  Unexpected duplicate selector ".alignfull", first used at line 180 (no-duplicate-selectors) [stylelint]

[14:00:31] Finished 'styles' after 1.84 s
```

ataylorme:
Also if I move `stylelint()` down after `AtImport` in the `postcssPlugins` array so it runs second a lot of warnings show up due to the value being imported into the CSS contents in the stream.

Example:

```sh
assets/css/src/widgets.css
6:1	⚠  Expected empty line before comment (comment-empty-line-before) [stylelint]
16:1	⚠  Expected empty line before rule (rule-empty-line-before) [stylelint]
21:1	⚠  Expected empty line before rule (rule-empty-line-before) [stylelint]
25:1	⚠  Expected empty line before rule (rule-empty-line-before) [stylelint]
38:1	⚠  Expected empty line before rule (rule-empty-line-before) [stylelint]
43:1	⚠  Expected empty line before comment (comment-empty-line-before) [stylelint]
48:1	⚠  Expected empty line before rule (rule-empty-line-before) [stylelint]
52:1	⚠  Expected empty line before comment (comment-empty-line-before) [stylelint]
[...]

PR #406, opened 2019-04-02
Fix duplicate selectors

benoitchantre:
## Description

Follow-up of #390 (related to #334). I missed some duplicate selectors after having resolved conflicts.

## List of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [ ] This pull request relates to a ticket.
- [x] This code is tested.
- [ ] This change has been added to CHANGELOG.md
- [x] I want my code added to WP Rig.

PR #408, opened 2019-04-02
Stylelint after postcss import

ataylorme:
## Description

Addresses issue #407. Runs the `stylelint` postCSS plugin after `postcss-import`.

## List of changes

## Checklist:
- [x] This code is tested.


ataylorme:
@benoitchantre I now think this PR fixes #407 as the duplicate styles addressed in #406 are reported by `stylelint`.

I do think duplicate selectors not being reported when `stylelint` is used in the same PostCSS run as `postcss-import` but I am not sure which tool it is a bug in (PostCSS, `gulp-postcss`, `styelint`, etc.). I will investigate that more and report the bug.

In the meantime, running PostCSS twice - once for imports and again for everything else - seems to have worked around the bug.

benoitchantre:
Thank you @ataylorme. Works for me.

Issue #430, opened 2019-04-18
Clearer documentation about naming of wprig folder in 2.0


ehowey:
## Issue Overview
In the documentation installing wprig v2.0 it states as follows:

1. Clone or download this repository to the themes folder of a WordPress site on your development environment.
  -DO NOT change the name of the theme directory from the default wprig.

In discussion during office time today I asked a question about this.  It turns out that renaming the folder is not such a big deal, the problem occurs when you rename the wprig folder to your production theme name which causes confusion in the "npm run bundle" process.

## Possible Solution

1. Clone or download this repository to the themes folder of a WordPress site on your development environment.
  -DO NOT give the wprig development folder the same name as your production theme. This will cause issues when you bundle your theme for production.  For instance if your theme will eventually be named "Excalibur" you cannot use this as your folder name.  The "excalibur" folder gets created during the "npm run bundle" process when you are ready to deploy.  See more detail on [the production process farther on in the readme.](https://github.com/wprig/wprig/blob/v2.0/README.md#how-to-build-wp-rig-for-production)

ataylorme:
@ehowey thank you for sharing that this wasn't clear for you. I think your suggested solution makes sense. Would you like to open a pull request to the README?

Issue #440, opened 2019-04-25
Menu Toggle Not Working Correctly


ColtonGO:
## Issue Overview

I am having the menu toggle issue previously reported. I just tried again with a fresh copy of wprig since the issue someone else created has been closed and moved to the done and it is still not working correctly for me. The only changes made to the fresh copy of wprig are to enable browsersync. Could this be something I didn't setup correctly? I followed along with the videos to install everything else required on my Windows 10 machine.
## Describe your environment

I am using VS Code, Google Chrome, Microsoft Edge, iphone6 (Google Chrome, Safari). I am using WAMP64 with a fresh copy of WordPress 5.1.1
## Steps to Reproduce

1. Sync all using browsersync
2. Menu toggle does not work on any browser on PC or iPhone 6 
3. Navigate away from homepage. Menu toggle works on all browsers except Microsoft Edge (PC)
4. Click website title and menu toggle doesn't work on any browsers.
5. Repeat steps 2-3 and sometimes it works in certain browsers, other times it doesn't.

## Expected Behavior

The menu toggle to display the menu when clicked
## Current Behavior

Clicking the site title seems to consistently break the menu toggle. Other times it's a toss up on whether it will work or not.
## Related Issues and/or PRs

Menu Toggle and Menu Dropdown not Working on iPad #289

benoitchantre:
Are you using the `v2.0` branch?
If yes, did you run `npm run build` before your tests?

ColtonGO:
Turns out I was using the master branch.
I downloaded v2.0 branch and followed the instructions within that branch for installation. I received the this error when running npm run dev:
`Requiring external module @babel/register
internal/modules/cjs/loader.js:711
    throw err;
    ^

SyntaxError: C:\wamp64\www\cressman\wp-content\themes\wprig-2.0\config\config.json: Unexpected token } in JSON at position 107
    at JSON.parse (<anonymous>)
    at Object.Module._extensions..json (internal/modules/cjs/loader.js:708:27)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
[...]

PR #427, opened 2019-04-15
Always copy, and run string replacement, on style.css

ataylorme:
## Description

Addresses issue #350. Always copy, and run string replacement, on `style.css`

Adds a new task `prodStringReplace`, which copies files requirted for the bundled theme but not created in a build step to the production theme and runs string replacement on them.

Rename `prodFinish` to `prodCompress` as the task creates the `.zip` of the production bundle.

Remove `style.css` from `filesToCopy` in the default config as it is now handled with the new `prodStringReplace` task.

## List of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] This pull request relates to a ticket.
- [x] This code is tested.
- [ ] This change has been added to CHANGELOG.md
- [x] I want my code added to WP Rig.

PR #441, opened 2019-04-25
Clarify naming of wprig development folder in the installation process

ehowey:
## Description
Addresses issue #430 
Clarification of the folder name restrictions when installing wprig.

## List of changes

## Checklist:
- [ ] This code is tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ WP Rig requires the following dependencies. Full installation instructions are p
 
 ### How to install WP Rig:
 1. Clone or download this repository to the themes folder of a WordPress site on your development environment.
-      - DO NOT give the wprig development folder the same name as your production theme. This will cause issues with the `npm run bundle` command during the production process. For instance if your theme will eventually be named "Excalibur" you cannot use this as your development folder name. The 'excalibur' folder gets created during the production process and cannot exist beforehand.
+      - DO NOT give the wprig development folder the same name as your production theme. This will cause issues with the `npm run bundle` command during the production process. For instance if your theme will eventually be named "Excalibur" you cannot use this as your development folder name. The 'excalibur' folder gets created during the production process and cannot exist beforehand. The default development folder name is `wprig`.
 2. Configure theme settings, including the theme slug and name.
     - View `./config/config.default.json` for the default settings.
     - Place custom theme settings in `./config/config.json` to override default settings.
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ WP Rig requires the following dependencies. Full installation instructions are p
 
 ### How to install WP Rig:
 1. Clone or download this repository to the themes folder of a WordPress site on your development environment.
-      - DO NOT give the wprig development folder the same name as your production theme. This will cause issues when you bundle your theme for production.  For instance if your theme will eventually be named "Excalibur" you cannot use this as your development folder name.  The 'excalibur' folder will get created during the 'npm run bundle' process when bundling the theme for production.  See more detail on [the production process farther on in the readme.](https://github.com/wprig/wprig/blob/v2.0/README.md#how-to-build-wp-rig-for-production) 
+      - DO NOT give the wprig development folder the same name as your production theme. This will cause issues with the 'npm run bundle' command during the production process. For instance if your theme will eventually be named "Excalibur" you cannot use this as your development folder name. The 'excalibur' folder gets created during the production process and cannot exist beforehand.  See more detail on [the production process farther on in the readme.](https://github.com/wprig/wprig/blob/v2.0/README.md#how-to-build-wp-rig-for-production) 
 2. Configure theme settings, including the theme slug and name.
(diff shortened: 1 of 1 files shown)


ataylorme:
@ehowey thank you for the pull request! Do you think it makes sense to add something like "for this reason, we suggest having the directory name of the WP Rig development theme be `wprig`".

ehowey:
I am hesitant about this...part of where this came up for me was in having multiple installs of wprig in one local wordpress install.  This started happening for me when I was flipping between v1.x and v2.0 - I got scared that something would break if I messed up the folder naming convention...I would have preferred to just do something like 'excalibur-dev' and then if needed 'excalibur-dev-2', etc. 

I also ran into an odd issue with npm run rig init when continuing to use the same folder name.  I should probably test this in a clean install and see more about what happened, if I recall correctly the rig init process didn't properly download all of the dependencies, it almost seemed as if it thought they were cached already or something.  What I did to make this happen was move my 'wprig' folder to a new directory (backed it up), and then download a fresh copy of 2.0 in folder 'wprig', ran `npm run rig init`.  This issue fixed for me if I renamed the folder to something like 'wprig3'...what I was doing then was renaming the folder, running the `npm run rig init` command, and then renaming the folder back to `wprig`.  

What if the sentence was "The default development folder name is `wprig`?

ataylorme:
Interesting, I hadn't considered folks having multiple copies of WP Rig. It makes sense during a migration period from `1.x` to `2.x`. Thank you for sharing more details about your usage.

benoitchantre:
> Interesting, I hadn't considered folks having multiple copies of WP Rig

Here's my setup: I have multiple copies of WP Rig (one for each theme based on WP Rig). They are located in an external directory where I have all my custom themes and plugins. Then, they are symlinked as `wprig` into the dev environment.

When I run `npm run bundle`, the bundle is located in `wp-content/themes/`. It works well.

ataylorme:
@benoitchantre are you using more than one WP Rig based theme per project? As WordPress can only have one theme active at a time, as opposed to many active plugins, my expectation is that there will be one WP Rig theme per project in most cases.

If there is a use case for actively developing two themes (or more) with WP Rig at the same time I would like to know more details.

GitaStreet:
@ataylorme 

"If there is a use case for actively developing two themes (or more) with WP Rig at the same time I would like to know more details."

I work on various projects, and also do testing and learning from various starter themes and frameworks. 

For work on localhost I use a WordPress Multisite installation. Just makes it easier and quicker to setup a subfolder for new tests. 

Also I sometimes found difficulties making the build process work with WP Rig - so it was relevant to keep the one that work - as well as try out the new versions as they appear :-)

benoitchantre:
> are you using more than one WP Rig based theme per project?

@ataylorme No. I setup a dev environment for each project (with Local by Flywheel). I work on client projects, but if I had to work on premium themes, I would probably adopt the same strategy.

In the case I had to keep many WP Rig projects inside the same environment, I would try (not tested) to use a name like `wprig-themeslug` for folder name, to avoid name collision.

ehowey:
Out of curiosity tonight I tried to reproduce the error I mentioned above with multiple iterations of the folder `wprig` in the same local wordpress install - it seemed similar to what was reported here today: https://github.com/wprig/wprig/issues/440

I couldn't get the error to happen again, at least on my first try, I basically did two fresh installs of wprig in a clean local dev environment.  Both in a folder named `wprig`.  It worked fine.  Whatever the bug was that we both encountered here must be more limited.

benoitchantre:
> Here's my setup: I have multiple copies of WP Rig (one for each theme based on WP Rig). They are located in an external directory where I have all my custom themes and plugins. Then, they are symlinked as wprig into the dev environment.
>
> When I run npm run bundle, the bundle is located in `wp-content/themes/`. It works well.

Not tested enough: it doesn't work, sorry. 🤦‍♂

Themes that are located in an external directory need a directory name like `wprig-themeslug` to prevent a name collision when you create the bundle with `npm run bundle`. If you use a `symlink` the bundle will be created your external directory, not in `wp-content/themes/`.

> In the case I had to keep many WP Rig projects inside the same environment, I would try (not tested) to use a name like `wprig-themeslug` for folder name, to avoid name collision.

It works, but you see multiple themes named `WP Rig`. #427 will solve that inconvenient.

benoitchantre:
> If there is a use case for actively developing two themes (or more) with WP Rig at the same time I would like to know more details.

I'm testing that: I have specific environment for each of my projects (client work) and an environment setup with the Theme Unit Test data.

Even if it is not complicated and long to create a new environment with the Theme Unit Test data, it's easier and faster to keep only one.

When I start to work on the theme, I start in the environment setup with the Theme Unit Test data. After some time, when common styles are defined, I work in the project environment where I have a content closer to the reality, additional plugins and the need for additional styles.

mor10:
I'm merging this. We can review as part of the documentation process.

Issue #444, opened 2019-04-29
Google Fonts: subset defaults to `latin-ext` instead of `latin`


benoitchantre:
## Issue Overview

`Latin Extended` allows to support more languages, but the file to download is bigger than `Latin`. If the website don't need extended characters, this is a waste of bandwidth.

By default Google Fonts gives you a link with `Latin` subset.

Theme authors should be able to change the character support.
The default value for `subset` should be `latin` because not all fonts support `latin-ext`.

Edit: looking at this issue again, I think the only change needed is an update of the default value for `subset`. Theme authors can change the value directly if needed.

mor10:
Is there a list somewhere of what languages use `latin-ext`?

benoitchantre:
[According to Wikipedia](https://en.m.wikipedia.org/wiki/Latin_Extended-A), languages that use the followings alphabets:

- Catalan
- Czec
- Latin
- Latvian
- Lithuanian
- Polish
- Slovak
- Turkish
- Welsh

I just found that Google Fonts serve a stylesheet that contains all sunsets to major browsers even if we specify a subset, because they [browsers are able download only what they need](https://caniuse.com/#feat=font-unicode-range) based on the Unicode ranges they find in the page.

See https://github.com/google/fonts/issues/1189

I think we can safely remove `subset` from the query arguments.

ataylorme:
Closed by #446

PR #446, opened 2019-04-29
Fix Google Fonts default subset

benoitchantre:
## Description

Addresses issue #444.

Remove the `subset` query arg to match the default value of Google Fonts (`latin`), adds inline documentation to change the default value.

## List of changes

## Checklist:
- [x] This code is tested.


mor10:
Apparently browsers are now smart enough to only download the font faces and extensions actually used in the current DOM, so this is safe.

Reference: https://github.com/google/fonts/issues/1189

Issue #449, opened 2019-04-30
Fatal Errors in when Sidebars component is not initialized


benoitchantre:
## Issue Overview

The method `is_primary_sidebar_active` is called in the _Styles_ component and in the _Image_Sizes_ component without a check if the method is available or not.

When the _Sidebars_ component is not initialized (because the theme doesn't use widgets) we get a fatal error.

benoitchantre:
Related to #515.

PR #450, opened 2019-04-30
Fix fatal errors in when Sidebars component is not initialized

benoitchantre:
## Description

Addresses issue #449.

Adds some checks to prevent fatal errors when the Sidebars component is not initialized.

## List of changes

## Checklist:
- [x] This code is tested.


inc/Image_Sizes/Component.php
```diff
@@ -51,7 +50,8 @@ public function filter_content_image_sizes_attr( string $sizes, array $size ) :
 			$sizes = '100vw';
 		}
 
-		if ( wp_rig()->is_primary_sidebar_active() ) {
+		if ( method_exists( 'wp_rig', 'is_primary_sidebar_active' )
```
felixarntz: I think this should be `method_exists( wp_rig(), 'is_primary_sidebar_active' )` right?

inc/Image_Sizes/Component.php
```diff
@@ -85,7 +85,8 @@ public function filter_header_image_tag( string $html, $header, array $attr ) :
 	public function filter_post_thumbnail_sizes_attr( array $attr, WP_Post $attachment, $size ) : array {
 		$attr['sizes'] = '100vw';
 
-		if ( wp_rig()->is_primary_sidebar_active() ) {
+		if ( method_exists( 'wp_rig', 'is_primary_sidebar_active' )
```
felixarntz: Same here.

inc/Styles/Component.php
```diff
@@ -312,6 +300,26 @@ protected function get_css_files() : array {
 			),
 		);
 
+		if ( method_exists( 'wp_rig', 'is_primary_sidebar_active' ) ) {
```
felixarntz: Same here.

sidebar.php
```diff
@@ -9,7 +9,8 @@
 
 namespace WP_Rig\WP_Rig;
 
-if ( ! wp_rig()->is_primary_sidebar_active() ) {
+if ( ! method_exists( 'wp_rig', 'is_primary_sidebar_active' )
```
felixarntz: Same here.

inc/Styles/Component.php
```diff
@@ -291,18 +291,6 @@ protected function get_css_files() : array {
...
-				'preload_callback' => function() {
-					return wp_rig()->is_primary_sidebar_active();
-				},
-			),
-			'wp-rig-widgets'    => array(
-				'file'             => 'widgets.min.css',
-				'preload_callback' => function() {
-					return wp_rig()->is_primary_sidebar_active();
-				},
-			),
```
felixarntz: Could we possibly just keep that bit and add the `method_exists()` checks inside the `preload_callback`? Or is there something I'm missing that wouldn't work then?

benoitchantre: If we only check inside the `preload_callback`, the styles will be enqueued when peloading styles is disabled. We don't need these files when there's no Sidebar component.

inc/Image_Sizes/Component.php
benoitchantre: It seems I have not tested my code properly: it doesn't work 🤦‍♂  
`method_exists( wp_rig(), 'is_primary_sidebar_active' )` doesn't work either.
`method_exists( 'WP_Rig\WP_Rig\Sidebars\Component', 'is_primary_sidebar_active' )` works, but only if the component has been completely removed (doesn't work if the component is simply not instanciated). I'll need some help to find the right solution.

benoitchantre:
@felixarntz PR is updated. I have added `template_tag_exists` method in `Template_Tags` class and used it instead of `method_exists`.

ataylorme:
@bamadesigner raised a good point in office hours that we haven't had a discussion on being able to remove components. The current codebase assumes all components will exist/be loaded.

ataylorme:
My opinion is that deleting unused components is not necessary. It will leave a bit more PHP but mostly actions/filters not being used if the corresponding feature isn't active (e.g. AMP or JetPack plugins not installed/activated).

In the case of the sidebar, you could simply keep the sidebar inactive and it won't render.

bamadesigner:
Hi @benoitchantre! Thanks for the PR and for joining our Office Hours earlier to discuss. Our core maintainers had a chat and decided we definitely want to explore a world where its easier in WP Rig to add/remove components but we want to give it due diligence so we created a separate issue and were going to work on this functionality in a future release.

You can view the issue here, if you'd like to comment or follow along: https://github.com/wprig/wprig/issues/515

Thanks again!

benoitchantre:
No problem, thank you for your comment @bamadesigner.

Issue #464, opened 2019-05-06
Nav_Menus component is not ready to handle many nav menus


benoitchantre:
## Issue Overview

The component seems limited to handle only one nav menu. I think it should ready to handle multiple nav menus, because a theme often has many nav menus.

WP Rig should be usable with many nav menus without having to refactor the methods of this component.

## Current Behavior

We have a `PRIMARY_NAV_MENU_SLUG` string to store nav menu slugs. This constant limits the usage of the component to one nav menu without having to refactor it.

## Possible Solution

Rename `PRIMARY_NAV_MENU_SLUG` to `NAV_MENUS` and store nav menus as an associative array with menu slugs and descriptif texts.

Refactor methods to use that array.

PR #465, opened 2019-05-06
Support multiple nav menus

benoitchantre:
## Description

Addresses issue #464.

Makes the component ready to store and register multiple nav menus with minimal changes.

When you declare a new menu location (eg. `footer`), you will need to add it in `set_navs_menus`.

```php
/**
 * Set the navs menus slugs and their description.
 *
 * @return void
 */
public function set_navs_menus() {
	$this->nav_menus = array(
		'primary' => esc_html__( 'Primary', 'wp-rig' ),
		'footer' => esc_html__( 'Footer', 'wp-rig' ),
	);
}
```

Then you will have to add related methods like this

```php
/**
 * Get the footer nav menu slug.
 *
 * @return string Footer menu location identifier (like a slug).
 */
public function get_footer_menu_slug() : string {
	return 'footer';
}

/**
 * Checks whether the footer navigation menu is active.
 *
 * @return bool True if the secondary navigation menu is active, false otherwise.
 */
public function is_footer_nav_menu_active() : bool {
	return (bool) has_nav_menu( $this->get_footer_menu_slug() );
}

/**
 * Displays the footer navigation menu.
 *
 * @param array $args Optional. Array of arguments. See `wp_nav_menu()` documentation for a list of supported
 *                    arguments.
 */
public function display_footer_nav_menu( array $args = array() ) {
	$args = wp_parse_args(
		$args,
		array(
			'container' => 'ul',

		)
	);

	$args['theme_location'] = $this->get_footer_menu_slug();

	wp_nav_menu( $args );
}
```

and expose the new methods to use them in the template files

```php
/**
 * Gets template tags to expose as methods on the Template_Tags class instance, accessible through `wp_rig()`.
 *
 * @return array Associative array of $method_name => $callback_info pairs. Each $callback_info must either be
 *               a callable or an array with key 'callable'. This approach is used to reserve the possibility of
 *               adding support for further arguments in the future.
 */
public function template_tags() : array {
	return array(
		'is_primary_nav_menu_active' => array( $this, 'is_primary_nav_menu_active' ),
		'display_primary_nav_menu'   => array( $this, 'display_primary_nav_menu' ),
		'is_footer_nav_menu_active'  => array( $this, 'is_footer_nav_menu_active' ),
		'display_footer_nav_menu'    => array( $this, 'display_footer_nav_menu' ),
	);
}
```

## List of changes

## Checklist:
- [x] This code is tested.


mor10:
Calling @bamadesigner

mor10:
I'm wondering if this falls under new feature (if so `future-release`).

inc/Nav_Menus/Component.php
```diff
@@ -26,7 +27,12 @@
  */
 class Component implements Component_Interface, Templating_Component_Interface {
 
-	const PRIMARY_NAV_MENU_SLUG = 'primary';
```
felixarntz: This must not be removed and should be reused, also since we have to work backward-compatible from now on.

inc/Nav_Menus/Component.php
```diff
@@ -59,15 +67,55 @@ public function template_tags() : array {
...
+	public function get_navs_menus() : array {
+		return $this->nav_menus;
+	}
+
+	/**
+	 * Get the navs menus slugs.
+	 *
+	 * @return array Array of menu location identifiers (like a slug).
+	 */
+	public function get_menus_slugs() : array {
```
felixarntz: I think `get_menu_slugs()` sounds better.

inc/Nav_Menus/Component.php
```diff
@@ -59,15 +67,55 @@ public function template_tags() : array {
...
+			'primary' => esc_html__( 'Primary', 'wp-rig' ),
+		);
+	}
+
+	/**
+	 * Get the navs menus slugs and their description.
+	 *
+	 * @return array Associative array of menu location identifiers (like a slug) and descriptive text.
+	 */
+	public function get_navs_menus() : array {
```
felixarntz: Same here, better call this `get_nav_menus()`.

inc/Nav_Menus/Component.php
```diff
@@ -59,15 +67,55 @@ public function template_tags() : array {
 		);
 	}
 
+	/**
+	 * Set the navs menus slugs and their description.
+	 *
+	 * @return void
+	 */
+	public function set_navs_menus() {
```
felixarntz: Same here, better call this `set_nav_menus()`. Also this should probably be private.

inc/Nav_Menus/Component.php
```diff
@@ -59,15 +67,55 @@ public function template_tags() : array {
...
+	public function get_menus_slugs() : array {
+		return array_keys( $this->nav_menus );
+	}
+
+	/**
+	 * Get the primary nav menu slug.
+	 *
+	 * @return string Primary menu location identifier (like a slug).
+	 */
+	public function get_primary_menu_slug() : string {
```
felixarntz: Let's remove this and stick with the constant as "always" on. That should be enforced for consistency.

inc/Nav_Menus/Component.php
```diff
@@ -59,15 +67,55 @@ public function template_tags() : array {
...
 	 */
 	public function action_register_nav_menus() {
-		register_nav_menus(
-			array(
-				static::PRIMARY_NAV_MENU_SLUG => esc_html__( 'Primary', 'wp-rig' ),
-			)
-		);
+		if ( empty( $this->nav_menus ) ) {
+			return;
+		}
```
felixarntz: Same here, I think this should not be checked. If you don't want nav menus at all, it should be encouraged to remove the entire component rather than having it still there without doing anything.

inc/Nav_Menus/Component.php
```diff
@@ -90,10 +138,15 @@ public function action_register_nav_menus() {
...
+
+		// Skip page menus (menus without location) and flat menus.
+		if ( empty( $args->theme_location ) || 1 === $args->depth ) {
+			return $item_output;
+		}
 
-		// Only for our primary menu location.
-		if ( empty( $args->theme_location ) || static::PRIMARY_NAV_MENU_SLUG !== $args->theme_location ) {
+		// Skip menus not assigned to our primary menu location.
+		if ( 'primary' !== $args->theme_location ) {
```
felixarntz: Let's make this use the constant again.

inc/Nav_Menus/Component.php
```diff
@@ -111,7 +164,7 @@ public function filter_primary_nav_menu_dropdown_symbol( string $item_output, WP
 	 * @return bool True if the primary navigation menu is active, false otherwise.
 	 */
 	public function is_primary_nav_menu_active() : bool {
-		return (bool) has_nav_menu( static::PRIMARY_NAV_MENU_SLUG );
+		return (bool) has_nav_menu( $this->get_primary_menu_slug() );
```
felixarntz: Same here.

inc/Nav_Menus/Component.php
```diff
@@ -121,11 +174,14 @@ public function is_primary_nav_menu_active() : bool {
...
 	public function display_primary_nav_menu( array $args = array() ) {
-		if ( ! isset( $args['container'] ) ) {
-			$args['container'] = 'ul';
-		}
+		$args = wp_parse_args(
+			$args,
+			array(
+				'container' => 'ul',
+			)
+		);
```
felixarntz: I think this is more complicated than the original code here.

inc/Nav_Menus/Component.php
```diff
@@ -121,11 +174,14 @@ public function is_primary_nav_menu_active() : bool {
...
 
-		$args['theme_location'] = static::PRIMARY_NAV_MENU_SLUG;
+		$args['theme_location'] = $this->get_primary_menu_slug();
```
felixarntz: Let's stick to the constant here.

benoitchantre:
Thank you for the review @felixarntz! I have updated the PR, let me know of more changes are required.

inc/Nav_Menus/Component.php
```diff
@@ -59,15 +69,40 @@ public function template_tags() : array {
...
+	public function get_nav_menus() : array {
+		return $this->nav_menus;
+	}
+
+	/**
+	 * Get the navs menus slugs.
+	 *
+	 * @return array Array of menu location identifiers (like a slug).
+	 */
+	public function get_menu_slugs() : array {
```
felixarntz: Revisiting this, I think this method isn't necessary. You could easily use `get_nav_menus()`. If you need to check for something using `isset()` is faster than `in_array()` anyway. :)

benoitchantre:
Thank you. Changes have been applied.

benoitchantre:
Ping @felixarntz

Issue #487, opened 2019-05-20
Harcoded colors similar or close to custom properties


benoitchantre:
## Issue Overview

Some colors are really close the global font color (`#333`). They should be either removed when they can inherit from the global font color or replaced the global font color facilitate.

https://github.com/wprig/wprig/blob/d355aef7c8372a7c5c979dc9dec75011946d211d/assets/css/src/global.css#L230

https://github.com/wprig/wprig/blob/d355aef7c8372a7c5c979dc9dec75011946d211d/assets/css/src/global.css#L528

For the main navigation links, it is difficult to see a difference between `#000`and `#333`.

https://github.com/wprig/wprig/blob/d355aef7c8372a7c5c979dc9dec75011946d211d/assets/css/src/global.css#L258

https://github.com/wprig/wprig/blob/d355aef7c8372a7c5c979dc9dec75011946d211d/assets/css/src/global.css#L389

PR #488, opened 2019-05-20
Reduce hardcoded colors

benoitchantre:
## Description

Addresses issue #487.

Replace hardcoded colors by existing and close custom properties.

## List of changes

## Checklist:
- [x] This code is tested.


assets/css/src/comments.css
```diff
@@ -159,7 +159,7 @@
 .comment-reply-link:focus,
 .comment-form .form-submit input:hover,
 .comment-form .form-submit input:focus {
-	border-color: #000;
+	border-color: var(--global-font-color);
```
felixarntz: For all the borders, I think it would be better to define a new variable `--global-border-color` and use that. It could use the same color like `--global-font-color` by default, but it'd be great so separate that.

benoitchantre: Thanks for the suggestion (applied).

assets/css/src/widgets.css
```diff
@@ -37,7 +37,7 @@
 
 /* Archive and Categories widgets */
 .widget_archive li {
-	color: #7d7d7d;
+	color: #808080;
```
felixarntz: I think having this is useless and weird. Why would we want to show this _very_ specific list in a different color? I think this rule should be removed entirely.

assets/css/src/_media.css
```diff
@@ -102,7 +102,7 @@ object {
 
 .gallery-item {
 	padding: 0.5em;
-	border: 1px solid rgb(222, 222, 222);
+	border: 1px solid #dedede;
```
felixarntz: This should be the `#ccc` color as well, it's so similar, and also a border. Related, my other comment below.

assets/css/src/comments.css
```diff
@@ -147,19 +147,19 @@
 	font-family: var(--highlight-font-family);
 	font-stretch: condensed;
 	font-size: 85%;
-	color: #000;
+	color: var(--global-font-color);
 	line-height: 1.3;
 	text-decoration: none;
 	background: transparent;
-	border: 1px solid #c3c3c3;
+	border: 1px solid #ccc;
```
felixarntz: I think for these `#ccc` borders we should also add a custom property, `--light-border-color`.

benoitchantre:
> You made a couple more changes that now lead me to another custom property - can you make this update as well? :)

Yes, I'll update this PR.

assets/css/src/comments.css
benoitchantre: Yes. I think `---border-color-light` would be more in line with what we have already.

assets/css/src/widgets.css
benoitchantre: > I think this rule should be removed entirely.

Yes. This widget displays a list of links, this color is never used (overriden by the link color)

assets/css/src/comments.css
benoitchantre: I think we should also rename `--global-border-color` to `--border-color-dark`. Ok for you @felixarntz?

benoitchantre:
@felixarntz Updated.

benoitchantre:
@felixarntz Can you review it again?

Issue #437, opened 2019-04-24
IE11 : sub menus don't work


benoitchantre:
## Issue Overview

We can see the following Javascript error located in `navigation.js`:
`Object doesn't support property or method 'forEach'`

https://github.com/wprig/wprig/blob/b3edfa4c488ff03aeed4770ed5a75973db4e3227/assets/js/src/navigation.js#L41-L43

Tested with IE11 in BrowserStack.

## Possible Solution

Replace the `forEach` loop by a `for` loop

ataylorme:
@benoitchantre I'm of the opinion that if your project requires IE 11 support you should add a polyfill. [Here is one for `forEach`](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Polyfill).

Alternatively, you could enqueue a new script [from polyfill.io](https://polyfill.io/v3/url-builder/) for the features you need to polyfill.

Perhaps this can be a documentation issue?

benoitchantre:
Thank you for the links @ataylorme.

If you don’t mind, I’m curious to knows why you advise to use a polyfill rather than to adapt the loop as we have only one usage of a `forEach` loop. From my readings, a `for` loop is a lot more performant than a `forEach` loop.

If WP rig doesn’t officially support IE11, I think that should be documented. Ideally, the documentation would include known issues and ways to solve these issues (like you answer)

ataylorme:
I am not against using a `for` loop in this specific case.

I guess my input for a polyfill is more a general opinion. Each WP Rig user's project needs will differ and someone supporting older browsers should look to use polyfills in their projects in general, regardless of it they are based on WP Rig

benoitchantre:
Thank you. I will submit a PR to fix that with a `for` loop.

Do you want I open a separate issue about a browser support documentation?

ataylorme:
Yes, a separate issue for browser support/IE documentation would be great. I think that is important as we move from beta to a full release

benoitchantre:
I did a PR that uses `for` loops instead of `forEach`. Unfortunately, sub menus are not working in IE11. Same result with a polyfill instead of this PR.

If someone has an idea why, I'm not against some help. 🆘

benoitchantre:
Closed by #438.

PR #513, opened 2019-06-12
Update navigation.js

idea--list:
I made a theme with WPRig v1.0.4 some months ago. On mobile devices the toggle menu has been unrelyabe and most of the time it did not work. In such cases i found a message in the console telling querySelector is undefined in navigation.js. It was strange as during development on localhost i could not reproduce it. As the website grew more complex, the more often i had to face the problem, so looked into navigation.js and found a solution that seems reliable for some weeks now.

## List of changes
- changed SITENAV back to a plain global var and defined it to be empty at the beginning
- assigned a value to it with the help of DOMContentLoaded event 

### DIFF
--- a/dev/js/navigation.js
+++ b/dev/js/navigation.js
@@ -5,13 +5,14 @@
  * navigation support for dropdown menus.
  */
 
-const SITENAV = document.querySelector( '.main-navigation' ),
-	KEYMAP = {
+var SITENAV;
+const KEYMAP = {
 		TAB: 9
 	};
 
 // Initiate the menus when the DOM loads.
 document.addEventListener( 'DOMContentLoaded', function() {
+	SITENAV = document.querySelector( '.main-navigation' );
 	initMainNavigation();
 	initMenuToggle();
 });


mor10:
Related: #473, #437

mor10:
@idea--list Have you tested the `v2.0` branch to see if the same issue persists?

idea--list:
The v2.0 branch i discovered just few hours ago and thus i did not yet test this PR in v2.0.
However i just checked the navigation.js of that version and inside that the SITENAV variable is not declared anywhere. So i am not sure if that version is affected by the same issue, but i will try the default v2.0 theme on the site overnight.

idea--list:
@mor10 
Just tried a theme generated with WPRig v2.0 on the site where i had that issue with the toggle menu before. The toggle menu works even without modifying any .js files of WPRig v2.0. The fix proposed in this PR is then only intended for v1.0.4

benoitchantre:
I think there was a similar issue in `v2.0` that has been fixed by #428.

jjloomis:
I ran into this as well (in v1.0.4.) I tweaked the [#428](https://github.com/wprig/wprig/pull/428) fix and worked it into my v.1.0.4 navigation.js file. It appears to be working. 

In my v1.0.4 navigation.js file, I got rid of this:
`// Initiate the menus when the DOM loads.
document.addEventListener( 'DOMContentLoaded', function() {
	initMainNavigation();
	initMenuToggle();
});`

and replaced it with this:

`if ( 'loading' === document.readyState ) {
	// The DOM has not yet been loaded.
	document.addEventListener( 'DOMContentLoaded', initNavigation );
} else {
	// The DOM has already been loaded.
	initNavigation();
}
// Initiate the menus when the DOM loads.
function initNavigation() {
	initMainNavigation();
	initMenuToggle();
}`

Note: If you're doing this, you have to change this part of the [#428](https://github.com/wprig/wprig/pull/428) fix:

`initNavToggleSubmenus();
initNavToggleSmall();`

to this (from v.1.04):

`initMainNavigation();
initMenuToggle();`

At least this is what's working for me (for now--haha.)

ataylorme:
@idea--list can you update the base of this pull request to be `v1-master` and not `master` (which is `v2`)?

idea--list:
After updating to WP 5.5 the unreliable behaviour of the mobile menu is back again, making my PR obsolete.
Downgrading to WP 5.4.2 solves the issue. So jquery or DOM state reporting might work different as of WP 5.5.

idea--list:
It seems line 15 of navigation.js is causing the issue as `document.addEventListener` seems to unreliable particularly on mobile browsers.

Changing line 15 to `window.onload = function() {`  solves seems to do the trick and fix the issue.

idea--list:
Updated my clients website to WP 5.7 and since that the menu is again more unreliable on mobile devices.
By that i mean about 2-3 times out of 10 nothing happens if i tap the menu button. This entraps the user on the last page they visited on the site. Sometimes it helps if one just reloads the page.

On desktop the issue does not occur. However all this is very annoying, i over the many months i tried several solutions but none of them seem to work in long terms...

Will this fundamental issue ever get solved in WP-Rig?

robruiz:
@idea--list - I would very much like to completely replace the default nav in WP Rig with something way better. I am willing to chip in, but not willing to do it alone. Are you interested in working together on this?

dthenley:
This is in reference to V1 of WP Rig

Issue #517, opened 2019-06-14
Footer does not always remain on the bottom of the page when there is minimal content


ehowey:
## Issue Overview
With the current CSS the footer is just placed at the end of the content, in almost all cases this results in the footer being at the bottom of the page or off the page entirely however on sites with smaller header areas and little content the footer comes up off the bottom of the page resulting in a less than appealing look.

## Describe your environment
N/A

## Steps to Reproduce
Create a page with a small amount of content, view it on a large resolution monitor.

## Expected Behavior
Footer to remain on the bottom of the page even when the content does not reach there.

## Current Behavior
Footer defaults to the end of the content.

## Possible Solution
See incoming PR, please review closely as there may be a better or more effective way to do what I did.

## Related Issues and/or PRs

#518

felixarntz:
See https://github.com/wprig/wprig/pull/518#issuecomment-504058621

PR #518, opened 2019-06-14
Always ensure the footer remains on the bottom of the page

ehowey:
## Description

Addresses issue #517 

Changes to the .site class to ensure the footer remains on the bottom of page even when there is minimal content.  Someone with some more CSS/development experience may want to do this differently.  This is just my proposed solution that seemed to work fine with the testing I could do.

## List of changes

## Checklist:
- [x] This code is tested.


felixarntz:
Thanks for opening the PR @ehowey!

What you're proposing is a rather opinionated change though. I understand often times theme developers would build their theme like that to cover for the case of short pages, but the way it currently is is the default behavior and how most websites work.

Since WP Rig is a starter theme and aims to be as unopinionated as possible, we shouldn't stray away from what the (modern) defaults of the web are and, particularly with styling, keep it simple.

ehowey:
Thanks @felixarntz - I am totally on board with the less is more approach when it comes to frameworks, bothers me when there is too much going on initially. This is something that can be easily handled by theme authors.  Just didn't know if it was intentional on your parts or not.

PR #536, opened 2019-06-21
Split block styles to separate files

benoitchantre:
## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/assets/css/src/_blocks.css
+++ b/assets/css/src/_blocks.css
@@ -1,11 +1,8 @@
-@import "_custom-properties.css";
-
 /*--------------------------------------------------------------
-## Blocks
+# Available block selectors:
 --------------------------------------------------------------*/
 
 /*
- * Available block selectors:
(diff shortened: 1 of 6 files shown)


benoitchantre:
This PR needs a bit more work to avoid duplication of styles for latest comments.

One solution could be use a `%` (silent) placeholder (requires to add postcss-extend).

benoitchantre:
We should not import the separate block files in `_blocks.css` but directly in `_content.css` and only if their usage is global.

Some blocks can be restrained to a specific post type with a template that will load a specific stylesheet. Blocks that are restricted to a specific template should not be part of a global stylesheet.

robruiz:
There are too many blocks and we don't want hundreds of css files in the theme. Going to close this. If theme devs want to do this , they are welcome to. However, we won't encourage this from the start. Comments in the one _blocks file should allow to break it up enough.

Issue #542, opened 2019-07-01
Missing PHP imports


benoitchantre:
## Issue Overview

Some components use external functions that are not imptorted with the `use` operator.

- AMP: `is_amp_endpoint`
- Customizer: `get_theme_file_path`
- Lazyload: `wp_allowed_protocols`, `esc_attr`
- Image_Sizes: `add_image_size`
- Styles: `add_editor_style`, `esc_html`

PR #543, opened 2019-07-01
Fix missing PHP imports

benoitchantre:
## Description

Addresses issue #542.

## List of changes

## Checklist:
- [x] This code is tested.


inc/AMP/Component.php
```diff
@@ -76,7 +77,7 @@ public function action_add_amp_support() {
 	 * @return bool Whether the AMP plugin is active and the current request is for an AMP endpoint.
 	 */
 	public function is_amp() : bool {
-		return function_exists( 'is_amp_endpoint' ) && \is_amp_endpoint();
+		return function_exists( 'is_amp_endpoint' ) && is_amp_endpoint();
```
felixarntz: I'd prefer if we didn't import this one because it's not from core and might not exist.

I know importing it without it existing wouldn't cause an actual issue, but I think it's good to make that separation between core functions (which always exist) and plugin functions (which might exist).

felixarntz (dismissed):
Thanks for these fixes. I have one minor suggestion.

inc/AMP/Component.php
benoitchantre: Ok. Updated following your advice.

szepeviktor:
Could we merge it?

benoitchantre:
@szepeviktor Pull requests need two approvals to be merged.
@felixarntz Could you check the latest changes?

robruiz:
@benoitchantre - how did you proceed with the requested change on this one? Was the requested change implemented on Sep 30th?

robruiz (approved):

Issue #567, opened 2019-08-05
Invalid value for `container` argument of `wp_nav_menu()`


felixarntz:
## Issue Overview
`wp_nav_menu()` supports a `container` argument that defines whether the menu's list element should be wrapped and, if so, how. The supported values are `div`, `nav` or `false` (to not provide a wrapping element). WP Rig currently automatically uses `ul`, which is an invalid value.

It doesn't actually break anything, because it will be interpreted as `false`, but is confusing and incorrect, so it should actually be set to `false`.

benoitchantre:
Closed by #569.

PR #569, opened 2019-08-05
Fix invalid nav menu argument

felixarntz:
## Description
Addresses issue #567 

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/inc/Nav_Menus/Component.php
+++ b/inc/Nav_Menus/Component.php
@@ -122,7 +122,7 @@ class Component implements Component_Interface, Templating_Component_Interface {
 	 */
 	public function display_primary_nav_menu( array $args = [] ) {
 		if ( ! isset( $args['container'] ) ) {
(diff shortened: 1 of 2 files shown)


inc/Nav_Menus/Component.php
```diff
@@ -122,7 +122,7 @@ public function is_primary_nav_menu_active() : bool {
 	 */
 	public function display_primary_nav_menu( array $args = [] ) {
 		if ( ! isset( $args['container'] ) ) {
-			$args['container'] = 'ul';
+			$args['container'] = false;
```
benoitchantre: From the documentation, [`$args['container']` should be a string](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/nav-menu-template.php#L28). When a container is not needed, an empty string should be passed. @felixarntz Let me know if I missed something.

Issue #414, opened 2019-04-09
Update lazyload to meet current best-practice


mor10:
Lazy loading of images (and iframes) is coming to a browser near you very soon. The lazy loading feature in WP Rig should be updated to meet this new standard: https://addyosmani.com/blog/lazy-loading/

Todo:

- [ ] - Add `loading="lazy"` attribute/value pair to all images to be lazy loaded.
- [ ] - Add `loading="eager"` to the first featured image on single posts and pages.
- [ ] - Update classnames to conform with fallback suggestion in @addyosmani's blog post (https://addyosmani.com/blog/lazy-loading/).
- [ ] - Remove the Customizer on/off feature as this will become a browser default.
- [ ] - Keep conditional for Jetpack for now.
- [ ] - Monitor https://github.com/joemcgill/WP_Media_Experiments for WP core developments.

Cc: @joemcgill

felixarntz:
I think this is better suited for post-2.0 because it's too early to implement this IMO. While we want to be on top of the time, implementing this now would be too much so. The feature is still in draft (see https://github.com/whatwg/html/pull/3752), Chrome is only including it experimentally.

In combination with this not being a breaking change, I suggest to punt to "Future Release".

mor10:
sure.

waviaei:
FYI. Lazy loading is now removed from Chrome in 76.

> The policy is removed in favor of a newer feature policy for loading, namely loading-frame-default-eager  which is more aligned with how the loading attribute will be used. The removal applies to both the Feature-Policy header and the <iframe> allow attribute.

https://developers.google.com/web/updates/2019/06/chrome-76-deps-rems

addyosmani:
Lazy loading has not been removed from Chrome 76. That update refers to an outdated feature policy (for controlling the default values for the feature), but the feature itself is still there.

waviaei:
@addyosmani Thank you for clarifying!

mor10:
@addyosmani (repeat from Twitter) what in your opinion is a good baseline for image loading in a CMS like WordPress? Setting all images to "lazy" and allowing devs to switch individual images to "auto" or "eager"? Or set all images to "auto" and allow devs to switch to "lazy"? Assume devs rarely change the default setting, and users stuff their pages full of images, mostly below the fold.

felixarntz:
Now that Chrome supports native lazy-loading, I created #577 to finally bring it to WP Rig (with graceful fallback to existing script).

mor10:
[...]

PR #577, opened 2019-08-08
Implement native lazy-loading, gracefully falling back to script

felixarntz:
## Description

Addresses issue #414 

Now that Chrome supports native lazy-loading, let's add this to WP Rig. The implementation in this PR is based on the suggestions from https://web.dev/native-lazy-loading:

* The existing lazyloading script is no longer enqueued from WordPress. It is still registered though, to maintain the integration with the service worker (from the PWA plugin) so that it can precache it.
* An inline script is added to the head which checks for support of the `loading` attribute.
    * If present, all values of relevant `data` attributes (as modified from the server-side) are rewritten back to their real attributes right away.
    * If not present, the original script for lazy-loading is inserted and works as it does today.

Another minor change I made (related): I added a note for the Customizer setting to toggle lazy-loading, that it will be ignored by AMP since AMP always lazy-loads.

cc @addyosmani @westonruter 

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/inc/Lazyload/Component.php
+++ b/inc/Lazyload/Component.php
@@ -72,7 +72,8 @@ class Component implements Component_Interface {
 		}
 
 		add_action( 'wp_head', [ $this, 'action_add_lazyload_filters' ], PHP_INT_MAX );
-		add_action( 'wp_enqueue_scripts', [ $this, 'action_enqueue_lazyload_assets' ] );
+		add_action( 'wp_enqueue_scripts', [ $this, 'action_register_lazyload_assets' ] );
+		add_action( 'wp_head', [ $this, 'action_print_lazyload_script' ], 8 );
 
 		// Do not lazy load avatar in admin bar.
 		add_action( 'admin_bar_menu', [ $this, 'action_remove_lazyload_filters' ], 0 );
@@ -130,10 +131,13 @@ class Component implements Component_Interface {
 	}
 
 	/**
-	 * Enqueues and defer lazy-loading JavaScript.
+	 * Registers and defers lazy-loading JavaScript.
+	 *
+	 * Although the script isn't directly enqueued via WordPress, it should still be registered so that it can be
+	 * precached by the service worker.
 	 */
-	public function action_enqueue_lazyload_assets() {
-		wp_enqueue_script(
+	public function action_register_lazyload_assets() {
+		wp_register_script(
 			'wp-rig-lazy-load-images',
 			get_theme_file_uri( '/assets/js/lazyload.min.js' ),
 			[],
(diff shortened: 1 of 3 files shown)


inc/Lazyload/Component.php
```diff
@@ -144,6 +154,44 @@ public function action_enqueue_lazyload_assets() {
...
+						if ( img.dataset.sizes ) {
+							img.sizes = img.dataset.sizes;
+						}
+					} );
+				} );
+			} else {
+				( function() {
+					var script = document.createElement( 'script' );
+					script.type = 'text/javascript';
+					script.src = '<?php echo esc_js( get_theme_file_uri( '/assets/js/lazyload.min.js' ) ); ?>';
```
swissspidy: This won't have the version number appended from `wp_rig()->get_asset_version( get_theme_file_path( '/assets/js/lazyload.min.js' ) )`, no?

felixarntz: Good catch. Not sure that's a problem though. Also some folks remove those suffixes with a filter, so it would be kind of hard to predict whether we should append it or not.

Now that's a fallback especially, it's unlikely this file will see updates.

inc/Lazyload/Component.php
```diff
@@ -144,6 +154,44 @@ public function action_enqueue_lazyload_assets() {
...
+	 * will be written back into their real attributes. If the browser does not support native lazy-loading, the script
+	 * bundled in the theme will be loaded.
+	 *
+	 * @see https://web.dev/native-lazy-loading
+	 */
+	public function action_print_lazyload_script() {
+		?>
+		<script type="text/javascript">
+			if ( 'loading' in HTMLImageElement.prototype ) {
+				document.addEventListener( 'DOMContentLoaded', function() {
```
westonruter: What about images loaded after the page has loaded, such as via an infinite scroll feature?

inc/Lazyload/Component.php
westonruter: The same filters could apply as when printing the script via WP_Scripts, then you'd know they are the same. Having the same URL would matter for precaching, I believe, unless the `ver` is stripped. I can't recall at the moment.

But anyway, if you could reference the previously-registerd script in WP_Scripts with the script_loader_src filter applied, then this will ensure consistency and it would allow the URL to be rerouted to another location potentially, such as when plugins provide a CDN for assets.

westonruter: Oh, and minor point: better to use `wp_encode_json()` here than use `esc_js()`, I believe

inc/Lazyload/Component.php
```diff
@@ -72,7 +72,8 @@ public function action_lazyload_images() {
 		}
 
 		add_action( 'wp_head', [ $this, 'action_add_lazyload_filters' ], PHP_INT_MAX );
-		add_action( 'wp_enqueue_scripts', [ $this, 'action_enqueue_lazyload_assets' ] );
+		add_action( 'wp_enqueue_scripts', [ $this, 'action_register_lazyload_assets' ] );
+		add_action( 'wp_head', [ $this, 'action_print_lazyload_script' ], 8 );
```
westonruter: This should be in `wp_print_footer_scripts` since it has to wait for DOM ready anyway. Better to not have the script block rendering in the head.

inc/Lazyload/Component.php
felixarntz: These are good suggestions, however I think this solution would add quite a bit of complexity not really worth it. I'll remove the version query string from that script entirely, this way we won't have to deal with it. It is highly unlikely the file will ever change anyway, especially now that it is a fallback.

inc/Lazyload/Component.php
felixarntz: Done via https://github.com/wprig/wprig/pull/577/commits/b888e3ce53cb41a4a170e48d4dd0e173bcf48292

inc/Lazyload/Component.php
felixarntz: I think that is something to revisit separately, as it's not covered by the current way the JS script works either.

inc/Lazyload/Component.php
```diff
@@ -130,20 +137,62 @@ public function action_add_lazyload_filters() {
...
+						if ( img.dataset.sizes ) {
+							img.sizes = img.dataset.sizes;
+						}
+					} );
+				} )();
+			} else {
+				( function() {
+					var script = document.createElement( 'script' );
+					script.type = 'text/javascript';
+					script.src = '<?php echo esc_js( get_theme_file_uri( '/assets/js/lazyload.min.js' ) ); ?>';
```
westonruter:
```suggestion
					script.src = <?php echo wp_json_encode( get_theme_file_uri( '/assets/js/lazyload.min.js' ) ); ?>;
```

inc/Lazyload/Component.php
westonruter: Good point! 👍

inc/Lazyload/Component.php
```diff
@@ -130,20 +137,62 @@ public function action_add_lazyload_filters() {
...
+	 * @see https://web.dev/native-lazy-loading
+	 */
+	public function action_print_lazyload_script() {
+		?>
+		<script type="text/javascript">
+			if ( 'loading' in HTMLImageElement.prototype ) {
+				( function() {
+					var images = [].slice.call( document.querySelectorAll( 'img.lazy' ) );
+					images.forEach( function( img ) {
+						img.src = img.dataset.src;
```
westonruter: I think this should potentially be moved to the bottom of the loop. Just thinking that some engines could potentially start requesting the `src` before the `srcset` is set, thus inadvertently downloading two images instead of one.

mor10:
This patch depends on whether Trac ticket [44427](https://core.trac.wordpress.org/ticket/44427) is merged into 5.3.

1. What do you think are the chances of that happening?
2. Will this patch cause issues with the current core proposal if implemented?

PR #588, opened 2019-08-25
Replace the dev theme directory with the prod theme directory when building for production.

ataylorme:
## Description

Addresses issue #559. Replaces the dev theme directory with the prod theme directory when building for production. Adds corresponding gulp tests for PHP, CSS and JS files.

## List of changes

## Checklist:
- [x] This code is tested.


ataylorme:
@benoitchantre can you review this one?

ataylorme:
>  It seems to me that this PR is trying to fix something caused by a bad practice

@benoitchantre that is the case. See my last comment in #559. However, I don't think doing this replacement harms anything. If you would rather close as `wontfix` due to trying to help bad practices I'm open to that.

benoitchantre:
> I don't think doing this replacement harms anything. If you would rather close as wontfix due to trying to help bad practices I'm open to that.

@ataylorme I think it is dangerous to do string replacement in path when directories are not renamed. This PR can create broken links when a directory inside the theme has the same name as the theme directory. String replacements will be done in the path but the directory will not be renamed. I know this is an edge case, but unless this PR also renames directories (is it a good idea and safe?), I’m in favor to close it because it can introduce that regression.

Issue #594, opened 2019-09-07
Generated zip file wrecks file/folder permissions!


justlevine:
## Issue Overview

Not sure what, why, or how to even debug this, but I just tried extracting that generated zip file on my server (siteground), and all the non-top-level files and folders had their permissions wrecked!

If i manually zip the dist theme, then extraction works fine.

## Describe your environment
Windows 10. WP-Rig 2.0.1, node 10.15.1, npm 6.4.1 to generate the bundled.zip, extracted to Siteground
## Steps to Reproduce

1. run `npm run bundle`
2. upload the generated zip to the wp-`content/themes` folder on the server.
3. extract the theme on the server.

## Expected Behavior
The file/folder permissions would be preserved
## Current Behavior
The permissions of all top-level files and folders are 666, and the permissions of all subfolders/files are nonexistent.

## Screenshots / Video
![Capture](https://user-images.githubusercontent.com/29322304/64478567-b6a28980-d1b2-11e9-8899-b060a8f1c8d9.PNG)

ataylorme:
@justlevine I'm not quite sure how to debug this either. We are using [`gulp-zip`](https://www.npmjs.com/package/gulp-zip) for creating the `.zip` archive.

You can disable the `.zip` creating in the configuration and create it manually as a workaround.

There is a [similar issue on that repo](https://github.com/sindresorhus/gulp-zip/issues/112) that suggests using `gulp-chmod` to change the permissions of the files prior to having them compressed.

However, I don't want to dictate file permissions to users of WP Rig. I am willing to help you modify your copy of WP Rig to do this. If it does indeed solve the problem, perhaps the file permissions can be a config setting.

justlevine:
Thanks! For now I'm manually zipping and uploading :)

I'm going to take a look at that GitHub issue and see if it resolves my problems, and report back. I will say that I've used gulp-zip in other projects, and never ran into this before...

Agree that file permissions shouldn't be dictated, and would humbly suggest holding off on that unless both a) the problem isn't straightforward to resolve and b) it's more widespread than just myself.

justlevine:
It seems to be a known issue with the zip folders explicitly created by gulp on windows machines (verses inherently).

Adding `gulp-chmod` worked (only the folder permissions needed to be changed), but so did adding `{nodir: true}` to all the gulp.src calls ( suggested [here](https://github.com/sindresorhus/gulp-zip/issues/64#issuecomment-191659258) ).

Not sure which approach is better.

[...]

PR #604, opened 2019-09-16
Use nodir:true in gulp src when creating the .zip

ataylorme:
## Description

Addresses issue #594. gulp `src` passes directories by default. However, on Windows and Linux the permissions are different so if the prod bundle is created on a Windows machine and deployed to a Linux server there are issues. Setting `nodir: true` in gulp `src` is a workaround suggested [here](sindresorhus/gulp-zip#64).

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/gulp/prodCompress.js
+++ b/gulp/prodCompress.js
@@ -29,7 +29,7 @@ export default function prodCompress( done ) {
 
 	return pump(
 		[
-			src( `${ prodThemePath }/**/*` ),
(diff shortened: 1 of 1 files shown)


mor10:
I was sortof able to test this. Works on Windows, and works on my weird Linux install on an RPI, but would be good to get it tested by someone who has a more legit Linux workflow.

PR #611, opened 2019-09-28
Fix invalid theme config causing issues with custom properties/media

felixarntz:
## Description
The default values for `custom-media` and `custom-properties` in `config.default.json` are not valid:
* They each should be an object with a `preserve` property (see `gulp/styles.js`).
* The `preserve` value for `custom-media` should default to `false` because basically _no_ browser today supports that.

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/config/config.default.json
+++ b/config/config.default.json
@@ -21,8 +21,12 @@
         "_custom-media.css"
       ],
       "features": {
-        "custom-media-queries": true,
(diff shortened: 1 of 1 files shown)


robruiz:
I'm interested to know if this is still a valid concern/contribution? I don't know enough about the context here to make a judgement.

PR #612, opened 2019-09-29
Simplify gallery style to be compatible with latest Gutenberg changes

felixarntz:
## Description
WordPress 5.3 will update the gallery block markup (via Gutenberg). All our adjustments currently break that, and I think the most straightforward solution is to simply remove our tweaks. The Gutenberg-provided styles look perfectly fine.

There are only two minor fixes required to make them work (both of which just resolve CSS specificity issues with other theme styles).

See [this core blog post](https://make.wordpress.org/core/2019/09/27/block-editor-theme-related-updates-in-wordpress-5-3/) for more information.

## List of changes

## Checklist:
- [x] This code is tested.

### DIFF
--- a/assets/css/src/_blocks.css
+++ b/assets/css/src/_blocks.css
@@ -109,87 +109,12 @@ ul.wp-block-latest-posts.is-grid.alignfull {
 # Gallery Block - overrides core styles
 --------------------------------------------------------------*/
 
-.wp-block-gallery {
-	display: grid;
-	grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
-	grid-gap: 0.5em;
+figure.wp-block-gallery {
+	margin-right: auto;
+	margin-left: auto;
 }
 
-.wp-block-gallery:not(.components-placeholder) {
-	margin: 1.5em auto;
-}
-
-/* Override standard ul indentation. */
-
-.entry-content .wp-block-gallery {
-	padding-right: 1.5em;
-	padding-left: 1.5em;
-}
-
-@media screen and (min-width: 48em) {
-
-	.entry-content .wp-block-gallery {
-		padding-right: 0;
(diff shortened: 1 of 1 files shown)


assets/css/src/_blocks.css
```diff
@@ -109,87 +109,12 @@ ul.wp-block-latest-posts.is-grid.alignfull {
...
-.wp-block-gallery.columns-2 .blocks-gallery-item,
-.wp-block-gallery.columns-3 .blocks-gallery-item,
-.wp-block-gallery.columns-4 .blocks-gallery-item,
-.wp-block-gallery.columns-5 .blocks-gallery-item,
-.wp-block-gallery.columns-6 .blocks-gallery-item,
-.wp-block-gallery.columns-7 .blocks-gallery-item,
-.wp-block-gallery.columns-8 .blocks-gallery-item,
-.wp-block-gallery.columns-9 .blocks-gallery-item {
-	grid-column: span 1;
+.entry-content .blocks-gallery-grid {
```
benoitchantre: It think it would be better to use `.wp-block-gallery .blocks-gallery-grid` instead of `.entry-content .blocks-gallery-grid` . The margin will be applied in other contexts than the post content.

benoitchantre: We should also keep some alignment adjustments (padding) for backward compatibility. Existing posts will contain the old gallery block markup until they are updated.

idea--list:
Just commented out all the overrides WP-Rig adds to the gallery related styles that ship with WP core. Now the gallery renders again as expected.
IMHO WP-rig should not add anything that is not absolutely needed as otherwise there are too many variables in a simple equation. I mean if WP core can change anything anytime and also WP-rig will begin to break something here and there without me having control over when all these might happen, then i will lose my mind :-)

PR #705, opened 2021-04-05
basic sprite support

cwrichardson:
## Description
Adds basic support for svg-sprites. Drop some SVG files into `assets/sprites/src` and during build it will output `assets/sprites/sprite.svg.php` as a `<symbol>`-style sprite which can be included with `get_template_part`, and `assets/sprites/sprite.css` which includes css to manage the dimensions of the sprite, and should be manually added to css files.

There's probably work to be done to make this "wprig ready", but putting this up so others can work on it. Here are my thoughts on things that could/should/must be changed before actually pulling into master:
- should add a fallback mode that creates PNGs for the sprites for ancient browser compatibility
- should have different outcomes for dev / build
  - for build, should include interim svg compiles (which is supported by svg-sprite library, I just haven't added it)
- svg-sprite supports a YAML file for inclusion of title and description and setting of an aria label for accessibility, but a) I couldn't get it to work; and b) even if I could have, I wasn't sure what I was going to do for translation
- probably it needs to do cleanup somewhere along the way

In addition to the above, the `get_template_part` seems ... like a gross hack to me, but it also seems to be the way this is done these days; so not sure if we should do it differently or not.

## List of changes
- updates dev requirements to pull in `gulp-svg-sprite`
- Adds a "sprite" action to `gulpfile.babel.js` and associated script in `gulp/`
- pull in any generated sprite into `header.php` with `get_template_part`

## Checklist:
- [ ] This code is tested.
- [ ] I want my code added to WP Rig.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@
 !/assets/images/
 /assets/images/*
 !/assets/images/src/
+!/assets/sprites
+/assets/sprites/*
+!/assets/sprites/src/
 !/languages/
 !gulp/
 !style.css
--- a/README.md
+++ b/README.md
@@ -32,6 +32,9 @@ WP Rig requires the following dependencies. Full installation instructions are p
 - [npm](https://www.npmjs.com/)
 - [Composer](https://getcomposer.org/) (installed globally)
 
+### WP Rig and child themes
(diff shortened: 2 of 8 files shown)


robruiz:
We would want you to merge this into the dev branch, not the master one. This feature, if pulled in, would be introduced in the next version of WP Rig

PR #768, opened 2021-11-19
added 'deps' to css files array

Spleeding1:
## Description

Addresses issue #

Added 'deps' array key to Styles/Component.php to allow for wp_enqueue_style dependencies.

## List of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:
- [X] This code is tested.

### DIFF
--- a/inc/Styles/Component.php
+++ b/inc/Styles/Component.php
@@ -306,7 +306,7 @@ class Component implements Component_Interface, Templating_Component_Interface {
 				},
 			),
 			'wp-rig-front-page' => array(
-				'file' => 'front-page.min.css',
+				'file'             => 'front-page.min.css',
 				'preload_callback' => function() {
 					global $template;
 					return 'front-page.php' === basename( $template );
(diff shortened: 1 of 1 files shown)


dthenley:
@Spleeding1 is there a good use case I can try this out with.

Spleeding1:
styles-b.css depends on styles-a.css, so styles-a.css has to load first

Spleeding1:
or probably better class in style-b.css overwrites style-a.css. It's built into WordPress enqueue but never implemented in WP Rig. This implements it.

dthenley:
After reviewing this and diving into the code a bit more it looks like dependencies for preloaded stylesheets doesnt exist just yet. When we preload the stylesheet is not traditionally enqueueing the stylesheet via wordpress `wp_enqueue_style` function rather it is creating a `<link>` element as seen here https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload.

Spleeding1:
It comes into play on lines 124 of Component.php

dthenley:
@Spleeding1 Haha I'm pretty foolish right now. Because i've been dealing with strictly wordpress and havent even evaluated how css files are included in an html document for a while here i was thinking css files were imported using `<style href="...">` and not `<link>` 🤦‍♂️. But I am still having issues with getting this to work with the style component. If you can come up with a use case so we can get it tested it out i think it would be a great addition to the code.

Spleeding1:
I will. Thanks for all of the work you are putting in. Believe me, I know what it is like to stare at something and have it be something stupid as to why it wasn't working

Issue #879, opened 2025-02-01
Corrupted files in bundled theme.


skywardpro:
**Describe the bug**
WPRig 3.0.1 - 3.0.2 corrupts files on bundle.

**To Reproduce**
1. Download lates version or WPRig and "npm run rig-init"
2. Go to bundled "wp-rig-test" theme folder
3. Try to open screenshot.png in the root
4. See that it's corrupted

**Expected behavior**
Files should be copied to bundled theme "as it is", without any changes and not being corrupted.

**Desktop (please complete the following information):**
 - Windows 11 Pro 24H2

**Additional context**
Same happens with additional files to copy (e.g. create custom folder with video or fonts in assets and try to copy it to bundle, it also appears corrupted).
But text files like readme or license which also copied as separate files are not corrupted. Seems it happens only with graphics.

<img width="855" alt="Image" src="https://github.com/user-attachments/assets/01c449ea-2ab1-4fb4-b727-1f5531587ed3" />

skywardpro:
I was able to fix it without relying on pump.
It's probably something with streams or how Gulp works with them in new version

Here is a quick solution for prodPred.js file (attached as .txt)

[prodPrep.js.txt](https://github.com/user-attachments/files/18629200/prodPrep.js.txt)

Due to its a "quickfix" I haven't opened pull request. 
I hope there should be another, more optimal solution and someone from contributors can fix it saving native logic.

skywardpro:
It was found that WPRIg brokes only files which are the part of filesToCopy list in theme config file e.g. screenshot.png in the root of theme, custom fonts and videos which have been added in by me in my custom wprig them.

I fixes it by changing the logic of copying these files without any minification and opened pull request: https://github.com/wprig/wprig/pull/882

PR #882, opened 2025-04-16
Update/prevent corruption for files to copy files

skywardpro:
## Description
After bundling some files become corrupted.
During the investigation it was found that:
1) Images copied via the dedicated images task (which processes assets/images/src/*) work correctly.
2) Files copied via the prodPrep task (which handles the filesToCopy array from your config, including screenshot.png) get corrupted.
The issue was found and described here: https://github.com/wprig/wprig/issues/879

Addresses issue #
My solution is to replace the Gulp stream logic in prodPrep.js with the manual Node.js fs stream logic. This bypasses the problematic Gulp handling specifically for the files listed in filesToCopy.

## List of changes
- [X] Bug fix: prevent corruption for images, fonts, videos files when copy them through the filesToCopy part in theme config file

## Checklist:
- [X] This pull request relates to a ticket.
- [X] This code is tested.
- [X] This change has been added to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
  - Fixed js minification and removes obsolete dev dependency
+ - Fixed corruption for files during the bundle for filesToCopy files
 
 ## 3.0.2
 
--- a/gulp/prodPrep.js
+++ b/gulp/prodPrep.js
@@ -4,11 +4,13 @@
 /**
  * External dependencies
  */
-import { src, dest } from 'gulp';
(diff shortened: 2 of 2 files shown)


robruiz:
@skywardpro - I already created a fix for this and it is in the develop branch. If could test that branch out and ensure the problem no longer exists on that branch, I would deeply appreciate it. Thank you.

robruiz (approved):
Looks like an interesting solve. Going to merge into develop and test there. Thank you.

Issue #884, opened 2025-07-27
dev mode crashes when php files are edited


erdmann040:
**Describe the error**
npm run dev sometimes crashes when editing PHP files, e.g. footer.php  

**How to reproduce the error**
1. Go to footer.php while $ npm run dev is running.
2. Insert a blank line after the opening footer tag.
3. Save the footer.php file.

This will trigger a warning (at least on my system)
```
[16:21:14] ‘watch’ reported an error after 33 seconds.
[16:21:14] Error: pump requires at least two streams.
    at pump (/home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/node_modules/pump/index.js:68:33)
    at FSWatcher.<anonymous> (file:///home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/gulp/watch.js:32:11)
    at FSWatcher.emit (node:events:530:35)
    at FSWatcher.emit (node:domain:552:15)
    at FSWatcher.emitWithAll (/home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/node_modules/chokidar/index.js:540:8)
    at FSWatcher._emit (/home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/node_modules/chokidar/index.js:632:8)
    at listener (/home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/node_modules/chokidar/lib/nodefs-handler.js:370:20)
[16:21:14] “Reload” is starting...
[16:21:14] “Reload” completed after 1.31 ms
[Browsersync] Browsers are reloading...
```

4. Remove the blank line you just added and save the file again

Now npm run dev crashes completely:
```
/home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/node_modules/pump/index.js:68
  if (streams.length < 2) throw new Error(‘pump requires two streams per minimum’)
                                ^
Error: pump requires at least two streams
    at pump (/home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/node_modules/pump/index.js:68:33)
    at FSWatcher.<anonymous> (file:///home/jan/lando/jgb/fewo/wordpress/wp-content/themes/wprig/gulp/watch.js:32:11)
    at FSWatcher.emit (node:events:530:35)
    at FSWatcher.emit (node:domain:552:15)
[...]

PR #885, opened 2025-07-27
Fix: No more npm crashes PHP file editing 

erdmann040:
Please see issue #884 (dev mode crashes when php files are edited)


erdmann040:
To be honest: Just a suggestion from Chat GPT—but it seems to work.

robruiz (approved):
Will test in develop. Looks good. Thank you.

PR #887, opened 2025-08-24
closes #875 (ES-Linter Problem)

### LINKED ISSUE #875: New npm task lint:js does not work correctly
**Describe the bug**
In version new npm commands have been added. While `npm run lint:css` works fine, `npm run lint:js` does not recognize formatting that does not comply with the WordPress coding standards for JS. 

**To Reproduce**
1. Execute `npm run lint:js`
2. Add the following line in a JavaScript file, whiche is not WPCS conform: 

```javascript
const apiRoot = `${window.location.origin}/wp-json/my-theme/v1/settings`;
```

**Expected behavior**
The linter should recognize that spaces are missing after the opening and before the closing curly bracket, but it does not.   

**Desktop (please complete the following information):**
 - OS: Fedora Linux
 - Code Editor: VS Code

**Reason for the error:**
An `eslint.config.js` was added in Version 3. If both an `eslint.config.js` and an `.eslintrc file` exist in the same project, then only the `eslint.config.js` is used, but the `.eslintrc` is ignored. Therefore, the WordPress linting plugin is no longer activated. 

**Solution:**
Because `eslint.config.js` is the more modern format, both configuration files should be merged into one `eslint.config.js` and the `.eslintrc` should be removed.

erdmann040:
Hi Rob, 

please take a critical look at the `eslint.config.js`. I had AI assistance and do not understand everything. However, it appears to be functioning.


robruiz:
@erdmann040 - in V3, we reworked the whole JS situation, lessened reliability on Gulp/babel for transpiling JS and going to a more esbuild driven solution. I would like to keep the dep list as low as possible. Not sure if that direction helps at all.

robruiz:
@erdmann040 - my guess here is that the eslint issues you all are trying to solve here stem from TS or JSX support? If so, then perhaps dep on Babel is unavoidable? Please let me know if that is the case.

erdmann040:
I didn't think about that at all. I was just happy to have found something that that worked. But I understand that you want to keep dependencies small and processing fast. 

I'll see if I can find a solution without babel dependencies in the next few days. 

But if there is WP Coding Standards for JSX and TS, we should try to support that too, right?

erdmann040:
Claude suggests the TypeScript parser as an alternative, which should be slightly faster and require fewer dependencies than Babel. However, there is no ESLint parser alternative that esbuild uses, as esbuild is a build tool and is not designed for the detailed AST analysis required for linting. (I have no idea what AST analysis actually is, but maybe you know more?)

Should we give TypeScript Parser a try?

Issue #758, opened 2021-10-08
Accessibility bug: Hidden site-title and site-description 


LukaszJaro:
Near the beginning of each page there is a hidden "home" link(site-title) and a site-description. These elements are not hidden from screen readers and are repeated on each page. This causes a lot of extra repeated chatter and the duplicate "home" link is focusable which causes a loss of focus for sighted keyboard only-users.

This code here appears to be adding this https://github.com/wprig/wprig/blob/master/template-parts/header/branding.php

If you tab through starting at the top, right after the logo, you will see a hidden item being focused on(or I should say the focus will disappear onto the hidden item), the hidden item is actually the site-title which has a link in it. The hidden link takes you back to the home page. It uses clip and position absolute to make it off-screen. The logo already takes you back to the homepage so this hidden link appears to be redundant.

If you run a screen reader, you will hear on every page the site-title and site-description being repeated.

You can see this in action on https://wprig.io/ or latest v2 site.

To sum up the issues:

- Loss of visible focus on the site-title
- Site-title has a back to homepage link in it which is redundant since logo already provides this link
- Extra chatter being repeated on each page via site-title and site-description

Recommendations:

1. Remove the hidden content. OR...
2. If this is required for SEO purposes, Remove the duplicate "home" link and use it's text as the alt text for the SureFund logo's image. Then hide the hidden site-description paragraph by adding the aria-hidden="true" attribute.

Something like this can be added to keep the SEO juice:

```
add_filter( 'get_custom_logo_image_attributes', 'add_custom_logo_image_attributes', 10, 3 );
function add_custom_logo_image_attributes( $custom_logo_attr ) {
	$custom_logo_attr['alt'] = get_bloginfo( 'name' );
	return $custom_logo_attr;
}
```

I looked further into how the custom logo function works and realized you technically don't need the filter I posted above. If the alt text on the logo image is empty, it actually adds the bloginfo automatically as you can see here:

```
if ( $unlink_homepage_logo && is_front_page() && ! is_paged() ) {
--
/*
* If on the home page, set the logo alt attribute to an empty string,
* as the image is decorative and doesn't need its purpose to be described.
*/
$custom_logo_attr['alt'] = '';
} else {
/*
* If the logo alt attribute is empty, get the site title and explicitly pass it
* to the attributes used by wp_get_attachment_image().
*/
$image_alt = get_post_meta( $custom_logo_id, '_wp_attachment_image_alt', true );
if ( empty( $image_alt ) ) {
[...]

PR #892, opened 2025-08-28
Fix(a11y): Prevent output of hidden site branding (#758)

SinghCod3r:
Resolves issue #758.

This pull request fixes an accessibility issue where the site title and description were visually hidden but still accessible to screen readers and keyboard navigation. This caused redundant links, loss of keyboard focus, and unnecessary screen reader chatter.

The fix updates template-parts/header/branding.php to wrap the output of the site title and description in a conditional check for the display_header_text Customizer setting. If a user has unchecked "Display Site Title and Tagline," the corresponding HTML is no longer rendered at all, providing a cleaner and more accessible user experience.


robruiz (approved):
This looks great! Thank you for your contribution!

Issue #904, opened 2025-09-05
React Warnings in Option Page


erdmann040:
**Describe the bug**
If you login and open the Theme Options Page in the backend and open the JS-Concol 3 Warning messagaes appear. 

**To Reproduce**
1. Setup Wprig as descipted in the Docs
2. Login and move the new Setings-Page
3. Open the Devtools and within the consol
4. You will see 3 warnig Messages

index.js:77 36px default size for wp.components.TextControl is deprecated since version 6.8 and will be removed in version 7.1. Note: Set the `__next40pxDefaultSize` prop to true to start opting into the new default size, which will become the default in a future version.

index.js:77 Bottom margin styles for wp.components.TextControl is deprecated since version 6.7 and will be removed in version 7.0. Note: Set the `__nextHasNoMarginBottom` prop to true to start opting into the new styles, which will become the default in a future version.

index.js:77 Bottom margin styles for wp.components.BaseControl is deprecated since version 6.7 and will be removed in version 7.0. Note: Set the `__nextHasNoMarginBottom` prop to true to start opting into the new styles, which will become the default in a future version.
﻿

**Expected behavior**
No Warnings 

**Screenshots**

<img width="1928" height="1164" alt="Image" src="https://github.com/user-attachments/assets/c91e1527-79ce-458a-a0fc-8ead6df4c0cc" />

Issue #899, opened 2025-09-04
Options Component does not work anymore


erdmann040:
**Describe the bug**
Only on develop branch 

If you activate the debug mode in the wp-config.php the following message shows up in the admin:

`Warning: filemtime(): stat failed for /app/wordpress/wp-content/themes/wprig/assets/js/admin/index.min.js in /app/wordpress/wp-content/themes/wprig/inc/Options/Component.php on line 64
`
 

**To Reproduce**
1. Setup wprig localy as documentend
2. Add to folling lines to the `wp-config.php`

```php
define( 'WP_DEBUG', true );
define( 'WP_DEBUG_LOG', true );
define( 'WP_DEBUG_DISPLAY', true );
define( 'SCRIPT_DEBUG', true );
```
3. Login in to the admin page
4. The folling message appear on the top of the page:

`Warning: filemtime(): stat failed for /app/wordpress/wp-content/themes/wprig/assets/js/admin/index.min.js in /app/wordpress/wp-content/themes/wprig/inc/Options/Component.php on line 64
`
The Options-Area wont work  anymore 

**Expected behavior**
The Options-Area should work and it should not show any error messages   
 
**Reason**
In one of my previous pull requests, I renamed `/assets/js/src/admin/index.js` to `index.jsx`.

It would probably be a violation of the WPCS to use the previous extension `js`.

It seems that the file is no longer minified because of the new extension, or the minified file cannot be found because of the new extention.

PR #905, opened 2025-09-05
Silence Gutenberg deprecation warnings on options page (#904)

erdmann040:
## Description

Addresses issue #904

This PR removes Gutenberg deprecation warnings on the options page by opting into the new component props and ensuring the tab content renders correctly. It does not change stored settings or runtime logic.

**Why it works**
- `__next40pxDefaultSize` opts `TextControl`/`SelectControl` into the new 40px control height.
- `__nextHasNoMarginBottom` opts controls into the margin-free styles.
- `ToggleControl` is used for toggles (supports the new props).
- `TabPanel` render callback uses `tab.name` (correct signature), so fields render reliably.

## List of changes

- Add `__next40pxDefaultSize` to `TextControl` and `SelectControl`
- Add `__nextHasNoMarginBottom` to `ToggleControl`, `TextControl`, and `SelectControl`
- Replace `FormToggle`/`BaseControl` combo with `ToggleControl`
- Fix `TabPanel` callback to use `tab.name`
- No functional or data changes; UI-only adjustments

## Checklist:
- [x] This pull request relates to a ticket. (Issue #904)
- [x] This code is tested. *(Manual: page renders as before; warnings no longer appear.)*

### DIFF
--- a/assets/js/src/admin/index.jsx
+++ b/assets/js/src/admin/index.jsx
@@ -5,9 +5,8 @@ import {
 	TabPanel,
 	TextControl,
 	SnackbarList,
-	BaseControl,
 	SelectControl,
-	FormToggle,
+	ToggleControl,
 } from '@wordpress/components';
 import { updateSettings } from './api.js';
 import formFieldsData from './settingsFields.json';
@@ -16,8 +15,8 @@ const textControlTypes = [
 	'text',
 	'email',
 	'url',
-	'password',
 	'number',
+	'password',
 	'search',
 	'tel',
 	'date',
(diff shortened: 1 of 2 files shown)


build-js.js
erdmann040, line 1: Cherrypicked my own pullrequest #899 here (without chages), so I could test the changes above.

assets/js/src/admin/index.jsx
erdmann040, line 1: Chat GPT suggests a lot of changes. These are not necessarily needed to solve the problem. If you want, I can submit a less invasive deletion. But the suggestions looked quite good.

robruiz: Please submit a PR with less invasive changes. Do you find Gulp being a problem here? Im starting to wonder if we would be better served just attempting to start 3.1 now and ditching Gulp altogether and recreating the bundle process with pure node scripts. Thoughts?

erdmann040: But maybe we should do this in 3.0.4. Version 3.0.2 has quite a few bugs and is not well documented. I think it would be good to release 3.0.3 quickly and postpone major refactorings for now.

Issue #908, opened 2025-09-06
Build fails: esbuild parses broken @wordpress/i18n sourcemap (build-module/index.js.map)


erdmann040:
### Summary
Running the JS build fails when esbuild tries to parse a **broken sourcemap** shipped by `@wordpress/i18n`. The error points to

`node_modules/@wordpress/i18n/build-module/index.js.map
`
As a result, `gulp build` aborts during the JS step.

### Affected versions
- WP Rig: 3.0.3 (develop)
- @wordpress/components: ^28.12.0 (dedupes i18n to 6.3.0)
but also in @wordpress/components: ^30.3.0 (dedupes i18n to 6.3.0)
- @wordpress/i18n: 6.3.0

### Steps to reproduce
1. Fresh clone, switch to `develop`.
2. `npm ci` (or `rm -rf node_modules package-lock.json && npm install`)
3. `npm run build`

### Actual behavior
`gulp build` fails at the JS build step:
✘ [ERROR] Expected end of file but found "c"

node_modules/@wordpress/i18n/build-module/index.js.map:1:713:
1 │ ...C,cAAc,QACR,gBAAgB","ignoreList":[]}c,QACR,gBAAgB","ignoreList":[]}
│ ^

bash
Code kopieren
This indicates the sourcemap file contains invalid JSON and esbuild aborts while resolving it.

### Expected behavior
The build should complete successfully. Vendor sourcemaps should be valid (or ignored safely) so bundling does not fail.

### Notes / Diagnosis
- The error originates from **reading the vendor sourcemap** referenced by `@wordpress/i18n/build-module/index.js` (via `//# sourceMappingURL=...`).
- Even after updating Gutenberg packages (e.g. `@wordpress/components` 30.x -> `@wordpress/i18n` 6.3.0), the sourcemap at the path above is still broken in our setup.

### Workarounds tested
- ✅ **Ignore the broken map during bundling**: add a tiny esbuild plugin that strips the `//# sourceMappingURL` line from `@wordpress/i18n/build-module/index.js`. This leaves the runtime code unchanged and the build succeeds.
- (Alternative) Pin a different `@wordpress/i18n` version via npm overrides or as a top-level devDependency. We’d prefer to avoid overrides if possible.
- (Alternative, larger change) Treat `@wordpress/*` as externals and rely on WordPress-provided handles (`wp-i18n`, etc.), so vendor sourcemaps from `node_modules` are never parsed.

### Proposed fix (small and scoped)
[...]

PR #909, opened 2025-09-06
Build/ignore i18n sourcemap

erdmann040:
## Description
Addresses issue #908

`gulp build` failed because esbuild tried to parse a broken sourcemap at:
`node_modules/@wordpress/i18n/build-module/index.js.map`.

This was first seen with `@wordpress/components@28.12.0` (resolving `@wordpress/i18n@5.26.0`) and still reproduced after upgrading to `@wordpress/components@30.3.0` (deduping to `@wordpress/i18n@6.3.0`) in our environment.

**Solution:** Add a tiny esbuild plugin that removes the `//# sourceMappingURL` line from `@wordpress/i18n/build-module/index.js` so esbuild does not attempt to read the faulty map. Runtime code remains unchanged.

## List of changes

**Details**
- `build-js.js`: introduce a small `stripI18nSourceMapPlugin` that strips the sourcemap directive only for `@wordpress/i18n/build-module/index.js` and run it before other plugins.

## Why this works
The failure occurred while esbuild followed `sourceMappingURL` and parsed a corrupt vendor map. By removing that single reference during bundling, the build completes successfully. The generated bundle is functionally identical; only the faulty vendor sourcemap is ignored.

## How to test
1. Clean install:
   ```bash
   rm -rf node_modules package-lock.json
   npm install
   ```
2. Build:
   ```bash
   npm run build
   ```
   ✅ No errors related to `@wordpress/i18n/build-module/index.js.map`.
3. (Optional) Verify that the plugin only targets the i18n ESM entry and that our own source maps still generate as usual.

## Checklist:
- [x] This pull request relates to ticket #908 
- [x] This code is tested. 

### DIFF
--- a/build-js.js
+++ b/build-js.js
@@ -58,6 +58,21 @@ const replaceInlineJSPlugin = {
 	},
 };
 
+// Tiny plugin: ignore the broken source map reference in @wordpress/i18n ESM entry
+const stripI18nSourceMapPlugin = {
+	name: 'stripI18nSourceMap',
+	setup( build ) {
+		// Cross-platform match for .../@wordpress/i18n/build-module/index.js
+		const filter = /@wordpress[\/\\]i18n[\/\\]build-module[\/\\]index\.js$/;
+		build.onLoad( { filter }, ( args ) => {
+			const code = readFileSync( args.path, 'utf8' )
+				// remove any `//# sourceMappingURL=...` so esbuild won't try to parse the broken map
+				.replace( /\/\/\s*#\s*sourceMappingURL=.*$/gm, '' );
+			return { contents: code, loader: 'js' };
+		} );
+	},
+};
+
 files.forEach( ( file ) => {
 	const relativePath = path.relative( srcDir, file );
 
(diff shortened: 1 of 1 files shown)


erdmann040:
Chat GPT suggests another alternative solution. I'm not sure if it would work. If so, it would probably be cleaner and faster.

Proposal: Treat @wordpress/* deps as externals in the JS build

### Why

Avoids bundling vendor code from node_modules, hence no broken third-party sourcemaps.

Smaller bundles; rely on WP-provided globals (best practice in the admin).

Lets us drop the temporary “strip broken sourcemap” plugin.

### What

Mark Gutenberg packages as external and map them to WP globals:

```
// build-js.js
import externalGlobal from 'esbuild-plugin-external-global';

const wpGlobals = {
  '@wordpress/i18n': 'wp.i18n',
  '@wordpress/element': 'wp.element',
  '@wordpress/components': 'wp.components',
  // add only the packages we actually import …
};

esbuild.build({
  // …
  format: 'iife',
  external: Object.keys(wpGlobals),
  plugins: [
    externalGlobal(wpGlobals),
    // remove stripI18nSourceMapPlugin after this change
    replaceInlineJSPlugin,
  ],
});
```
Enqueue with Core script handles so WP provides the globals:

```
wp_enqueue_script(
  'wprig-admin',
  $uri,
  [ 'wp-i18n', 'wp-element', 'wp-components' ], // match what we import
  $version,
  true
);

```

### Result
* Build succeeds without touching vendor sourcemaps.
* No runtime change (WP already ships these scripts).
* We can remove the sourcemap strip plugin once this lands.

PR #914, opened 2025-09-06
chore(lint): reduce linting issues in gulp files; unify formatting & apply autofixes 

erdmann040:
- Apply **uniform formatting** and ESLint **autofixes** across gulp files.
- Fix variable shadowing, early unused assignments, and missing returns.
- Keep current TypeScript parser/config; **no new deps**.
- Some errors/warnings remain to avoid risky changes.
- Verified: `npm run build` and `npm bundle` do not display any error messages.


erdmann040:
The gulp/editorStyles.js uses dependencies such as postCSS. I assume these can be deleted?

Issue #919, opened 2025-09-07
CSS Sourcemaps generated but not linked


erdmann040:
**Describe the bug**

On calling `npm run dev` sourcemaps are generated, but not linked in the unminified `*.min.css`.

**To Reproduce**
1. Install WPrig as documented 
2. Use the following `config/config.local.json`, which should enable linting for scripts an css, right?: 
```json
{
  "dev": {
    "styles": {
      "stage": 3,
      "autoprefixer": { "grid": false },
      "importFrom": ["_custom-media.css"],
      "features": {
        "custom-media-queries": true,
        "custom-properties": true,
        "nesting-rules": true
      },
      "sourceMaps": "inline"
    },
    "scripts": {
      "sourceMaps": true
    },
    "debug": {
      "styles": true,
      "scripts": true,
      "phpcs": true
    }
  }
}
```

@robruiz By the way: In my opinion, sourcemaps are a must and should at least show up in config/config.default.json so the user can see how to turn them on if needed.

3. `npm run dev`
4. No sourcemap links add the end of the unminfied `*.min.css`

PR #921, opened 2025-09-10
Fix/919 css sourcemaps v3

erdmann040:
## Description
Addresses issue #919

Sourcemaps for compiled CSS pointed to the top-level importer instead of the actual source files. In DevTools this made it hard to jump to the rule’s true origin. Additionally, configuration-driven preloads (custom media, etc.) had to be injected without breaking `@import` ordering.

**What this PR does**
- Switches the CSS pipeline to LightningCSS `bundleAsync()` so `@import` trees are resolved correctly and sourcemaps retain original sources.
- Introduces a virtual preload import that concatenates configured preload files once and inserts them *after* the top-level `@import` block. This preserves CSS ordering while keeping the preloads effective.
- Derives LightningCSS `targets` from the project Browserslist (env-aware via `BROWSERSLIST_ENV`), ensuring output matches the supported browsers.
- Renames `dev.styles.importFrom` → `dev.styles.preload` (clearer naming); keeps backward compatibility by still honoring `importFrom` with a deprecation warning.
- Continues to read merged configuration via `themeConfig` so overrides from `config.json` / `config.local.json` are respected.

**Why this works**
- Bundling with LightningCSS preserves file identity in the map; DevTools can now navigate to the correct original file & line.
- The virtual preload avoids mutating real sources and prevents late `@import` violations while still applying project preloads consistently.
- Browserslist-driven targets keep transpilation in sync with the project’s supported environments.

## List of changes

### Detailed changes
- Switched from `transform()` to `bundleAsync()` to preserve file identity in sourcemaps.
- Added virtual import mechanism so preloaded CSS files (e.g. `_custom-media.css`) appear correctly without breaking `@import` order.
- Introduced `dev.styles.preload` in config. Old `importFrom` still works but triggers a deprecation warning.
- Removed temporary debug logging.
- Browser targets now pulled dynamically from Browserslist config instead of hardcoding.

## How to test

1. **Build assets**
   ```bash
   npm run build:css
   ```
   - Verify that `assets/css/*.min.css` and `assets/css/editor/*.min.css` are generated.
   - Check that each has a `.map` file and ends with a `/*# sourceMappingURL=... */` comment.

2. **Verify sourcemaps in DevTools**
   - Run `npm run dev` and open the site in a browser.
   - In DevTools, inspect a style (e.g. from `global.css`).
   - **Expected:** The “jump to definition” goes to the correct partial (`_typography.css`, `_elements.css`, etc.), not the top-level importer.

3. **Check preload handling**
   - If the main menu works on both small and large viewports, the `_custom-media.css` is available in compiled outputs.   Maybe check also, that the preload-config works with other files as well.

4. **Check Browserslist targets**
   - Run two builds with different environments:
```
     //.browserslistrc
    > 1%
    last 2 versions
    
    [modern]
    last 1 chrome version
    last 1 firefox version
    last 1 safari version
    
    [legacy]
    ios_saf 12
    safari 12
``` 

```bash
     BROWSERSLIST_ENV=modern npm run build:css && cp assets/css/global.min.css global.modern.css
     BROWSERSLIST_ENV=legacy npm run build:css && cp assets/css/global.min.css global.legacy.css
     diff -u global.modern.css global.legacy.css | head -n 80
```
   - **Expected:** Differences in features like media query syntax (`min-width` vs. `width>=`), flexbox/grid prefixes, etc.

5. **Deprecated config key**
   - Temporarily set `dev.styles.importFrom` in `config.local.json` (instead of `preload`).
   - Run the build.
   - **Expected:** It still works, but a warning is logged:  
     `⚠️  dev.styles.importFrom is deprecated, please use dev.styles.preload instead.`

## Checklist:
- [x] This pull request relates to a ticket (#919).
- [x] This code is tested (manual build and inspection).


erdmann040:
@robruiz ChatGPT was very helpful to me here. I believe there is still room for improvement in the code. A large part of the complexity seems to come from preloading _custom-media.css. Is this actually necessary? I vaguely remember that Custom Media Queries have some special role, but I can’t recall exactly what it was or why.

If preloading is not required, we might want to consider simply importing _custom-media.css where it’s needed. That could potentially simplify the code quite a bit.

I also find the non-minified *.min.css files and the separate maps a bit confusing, even though I’m sure there were good reasons for this setup. I wonder if it might make sense to make that configurable.

